PR #21, opened 2018-04-10
Refactor gulp and format files

michaelmccombie:
- Refactored Gulp process
- Added CSS mapping
- Concatenated, minified, and formatted CSS
- Concatenated, minified, and formatted JS


.editorconfig
```diff
@@ -1,53 +0,0 @@
-# EditorConfig is awesome: http://EditorConfig.org
```
alexgibson: Editor config is something we use in all our projects, to help keep consistency with things like line spacing and indentation. Let's not delete it here, and I'd encourage everyone to [configure your editor](http://EditorConfig.org) to use it. It helps save time in code review and keeps things consistent 👍

.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules
 dist
 .netlify
+bower_components/
```
alexgibson: Unless we have a really good reason to support it, I'd suggest we not publish this to Bower. While it's maintained as far as bug fixes go, even the maintainers [try to enourage npm/yarn](https://bower.io/) these days. It's unlikely to get many new features, and is already way behind npm.

bower.json
```diff
@@ -0,0 +1,29 @@
+{
```
alexgibson: Again, I'd suggest we not publish to bower as it's largely unmaintained.

gulp/config.js
```diff
@@ -1,11 +1,66 @@
-'use strict';
+var dest = "./dist/assets/";
```
alexgibson: There's a good mixture of both ES5 and ES6 in this PR. 

For Node.js code, we can freely use ES6, so things like `var` can almost never be used (favoring `let` and `const` instead), and we can also use things like arrow functions to make code easier to read. I'd suggest we pick one version of thelanguage and stick to it. I'd favor ES6 for all node.js related code as it's easier to read, but we can use whatever people feel most comfortable with.

We can open a new issue to add ESLint to the repo, which will help enforce a standard.

gulp/config.js
```diff
@@ -1,11 +1,66 @@
...
+var src = "./src/assets";
+var version = "/1.0.0";
 
 module.exports = {
+  concatJS: {
```
alexgibson: There's a mixture of 4 space and 2 space indentation in these files. We favor 4 spaces for JS/CSS, and 2 spaces for HTML. Using editorconfig (as mentioned earlier) will help your editor follow this rule automatically.

gulp/config.js
```diff
@@ -1,11 +1,66 @@
...
+    documentation: {
+      src: [
+        src + '/js/site/global.js',
+        src + '/js/site/prism.js'
+      ],
+      dest: dest + '/js'
+    },
+    protocol: {
+      src: [
+        src + '/js/protocol/protocol-accordion.js',
```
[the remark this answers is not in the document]
alexgibson: Is concating JS files like this something we'd want? I'd imagine on bedrock, we'd want to handle the bundling ourselves, so only pages that had an accordion contained `protocol-accordion.js` for example. Bundling everything together in one big glob on JS could mean a lot of unused code?

gulp/config.js
```diff
@@ -1,11 +1,66 @@
...
+  },
+  sass: {
+    src: [
+      src + '/sass/protocol/protocol.scss',
+      src + '/sass/site/site.scss',
+      src + '/sass/site/prism.scss'
+    ],
+    dest: dest + '/css/'
+  },
+  uglify: {
```
alexgibson: Is uglify something we should handle in this project, or leave to sites that consume Protocol? (I could lean either way here, but perhaps something worth a discussion).

gulp/config.js
alexgibson: P.S - we can always fix this up in a followup PR 👍

alexgibson:
P.S - I'd be happy to help work on ESLint and CI config here if it helps. We could always address this after the PR here merges.

gulp/tasks/concatJS.js
```diff
@@ -0,0 +1,18 @@
+var gulp = require('gulp');
+var concat = require('gulp-concat');
+var plumber = require('gulp-plumber');
+var handleErrors = require('../utils/handleErrors');
+var config = require('../config').concatJS;
+
+gulp.task('concatJS', ['cleanCss'], function() {
```
alexgibson: This is a nit, but typing camelCase on the commandline makes my eye twitch 😉 . Can we stick to some sort of consistency using lowercase? A common convention for gulp is to use something like `gulp css:clean` or `glup css:lint` etc..

gulp/config.js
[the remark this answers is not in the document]
michaelmccombie: @craigcook that's a good catch, I was running through things quickly just to make sure everything is working but I can refactor this so we don't have to manually add each entry.

@alexgibson It makes sense to use javascript on a per instance basis in bedrock, I think we should do that. But I also think there is a benefit in having a single downloadable js file for anyone using Protocol outside of our team and outside of bedrock. So I don't see much harm in leaving it in for the time being.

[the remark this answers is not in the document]
alexgibson: As long as the build script provides both options, that works for me 👍

gulp/config.js
```diff
@@ -16,25 +71,12 @@ module.exports = {
...
-
   watch: {
     watchers: [
       {
         match: ['./src/static/**/*'],
         tasks: ['copy']
       },
-      {
-        match: ['src/assets/sass/**/*.scss'],
-        tasks: ['css:lint']
```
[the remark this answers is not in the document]
michaelmccombie: Good call, added that back in.

gulpfile.js
```diff
@@ -1,96 +1,3 @@
...
-// Register frontend composite task
-gulp.task('frontend', [
-  'drizzle',
-  'copy',
-  'sass',
-  'css:lint',
-  'js'
-]);
-
-// Register build task (for continuous deployment via Netflify)
```
alexgibson: Removing this task is going to break [Netlify deployments](https://github.com/mozilla/protocol/blob/master/netlify.toml#L3)

https://app.netlify.com/sites/mozilla-protocol/deploys/5acf879c4ed62f4a8757c33a

.editorconfig
michaelmccombie: This is good to know 👍, I was rolling back some of the linting work that Craig had done previously and didn't quite understand what this file was doing. I added it back in.

.gitignore
michaelmccombie: This makes sense. I worked on projects before where we used both Bower and NPM (with the delineation being that Bower is for front-end dependencies and NPM is for back-end dependencies) but I haven't kept up with any of it. We should definitely remove Bower and maintain everything with NPM.

bower.json
michaelmccombie: 👍

gulp/config.js
michaelmccombie: Yeah, there was definitely some copying and pasting going on from previous projects I've worked on without paying much attention to ES6 syntax.

gulpfile.js
michaelmccombie: Good catch! Just added this back in: https://github.com/mozilla/protocol/pull/21/commits/3a646fea8f76505e836464f550d8add122f7e059

gulp/tasks/default.js
```diff
@@ -0,0 +1,41 @@
...
+[
+  'clean',
+  'copy',
+  'js',
+  'serve',
+  'watch'
+].forEach(name => tasks[name](gulp, config[name]));
+
+// Register frontend composite task
+gulp.task('frontend', [
```
craigcook: We need to add `copy` to this task list for the static files.

gulp/config.js
```diff
@@ -1,11 +1,65 @@
...
+    src: src + '/sass/protocol/protocol.scss',
+    dest: dest + '/css',
+    base: "../../../../",
+    replacement: "../../../bower_components/"
+  },
+  sass: {
+    src: [
+      src + '/sass/protocol/protocol.scss',
+      src + '/sass/site/site.scss',
+      src + '/sass/site/prism.scss'
```
craigcook: Still missing out on demo styles here. And I still think it makes sense to process the whole folder and not individual files. I think it just means switching this to `src + '/sass/**/*.scss'` (and updating a few similar paths in other places).

gulp/config.js
```diff
@@ -1,11 +1,65 @@
...
+      src: [
+        src + '/js/protocol/*.js'
+      ],
+      dest: dest + '/js'
+    }
+  },
+  lintCss: {
+    src: [
+      dest + "/sass/protocol/protocol.css",
+      dest + "/sass/site/site.css"
```
craigcook: I think these should be `.scss` files, not `.css`? And this would be another place to do the whole folder like `"/sass/**/*.scss"`

gulp/config.js
```diff
@@ -1,11 +1,65 @@
...
+    ]
+  },
+  cleanCss: {
+    src: [
+      dest + '/css/protocol.css',
+      dest + '/css/site.css',
+      dest + '/css/prism.css'
```
craigcook: Process the folder as `dest + '/css/**/*.css'`?

gulp/config.js
michaelmccombie: Switching this to `src + '/sass/**/*.scss` is causing the build to generate duplicated scss files. So I included the demos scss file and I'll create a task to come back to this issue.
image

PR #28, opened 2018-04-13
Configure ESLint, Stylelint and run on CI

alexgibson:
- Adds ESLint config & gulp task.
- Fixes Stylelint task which was previously broken (I've removed the postcss bits for now, as we probably want a separate post-build task if we also want to lint compiled CSS?)
- Enables linting & build in CI on pull requests.
- Moves Stylelint rules to `stylelint.config.js` in the root directory, so text editors can pick up the definitions and provide auto-linting as you type 🔥 
- Updates Stylelint rules to match bedrock.


alexgibson:
@craigcook OK, I have CI running! I've gone and fixed the JS linting errors, but I haven't done the CSS errors just yet. Wanted to double-check with you that the Stylelint rules currently in the repo are all good? (I'm happy to follow them, just wanted to confirm).

alexgibson:
Ok, build is passing! This should lint front end code & compile the site against every pull request once merged.

Issue #39, opened 2018-04-17
js:uglify and css:clean Gulp tasks do not work


alexgibson:
- Both the `js:uglify` and `css:clean` gulp tasks are currently misconfigured and fail silently.

michaelmccombie:
I resolved the css:clean issue and included it as a commit https://github.com/mozilla/protocol/pull/40/commits/18fcab4d3e333e697a5f392b31d949d71e70ca93 in this branch https://github.com/mozilla/protocol/pull/40

Uglify is something that we wanted to discuss at some point https://github.com/mozilla/protocol/issues/25 . We can either remove it now or fix the bug by editing the path names as shown below.

<img width="457" alt="screen shot 2018-04-17 at 6 04 25 pm" src="https://user-images.githubusercontent.com/5924575/38906175-dded2a06-4269-11e8-9559-6d0da35d1434.png">

Issue #38, opened 2018-04-17
Provide non-concatenated JS files in the build


alexgibson:
Currently the build process concatenates JS files into a single `protocol.js` bundle. Whilst this may be ok for some sites that don't have their own asset pipeline, so sites like bedrock we probably want to cherry-pick which JS libs to include on a per page basis. We should ensure individual JS files are copied across to `./dist/assrts/js/`, as well as the bundle.

Issue #24, opened 2018-04-12
Remove Bower and replace Bower dependencies


michaelmccombie:
Per Alex's comment - "Unless we have a really good reason to support it, I'd suggest we not publish this to Bower. While it's maintained as far as bug fixes go, even the maintainers try to enourage npm/yarn these days. It's unlikely to get many new features, and is already way behind npm."

PR #40, opened 2018-04-17
Refactored gulp to process assets for publishing to NPM

alexgibson:
- Process protocol and documentation assets separately:
  - Protocol assets are copied to `./dist/assets/protocol/{css|js}/`
  - Docs assets are copied to `./dist/assets/docs/{css|js}`
- Fix JS and CSS compression tasks (Fixes https://github.com/mozilla/protocol/issues/39).
- Makes sure unconcatenated JS files are copied over (Fixes https://github.com/mozilla/protocol/issues/38).
- Remove Bower support (Fixes https://github.com/mozilla/protocol/issues/24)


alexgibson:
OK, after talking with Paul re: the bedrock side I have some small modifications I'd like to make. Putting this back to WIP for now until it's ready for review.

michaelmccombie:
There was an error (that also exists in master) with the path referenced in the css:clean gulp task. So I made a commit on this branch to fix that. Other than that everything looks good so far. I'll take another look when you are finished making the modifications you discussed with Paul.

alexgibson:
> There was an error (that also exists in master) with the path referenced in the css:clean gulp task. So I made a commit on this branch to fix that.

Thanks for the commit, but I think this is only part of the issue. The task works independently via `gulp css:clean`, but not when run as part of `gulp build`. I think the main issue is we're running a some tasks in parallel, which should be happening in serial. I'll see if I can try and fix it in this PR.

alexgibson:
@michaelmccombie OK I think everything should be ready for review. I fixed both the JS and CSS compression tasks and made sure all tasks `return` their streams correctly, so they are guaranteed to excecute in order 👍

PR #42, opened 2018-04-19
Update package.json

michaelmccombie:
This PR updates the package.json to point to our "main" file so we can publish to NPM. It also removes a Bug Fixes reference which was accidentally carried over from our design-tokens repo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,5 @@
 # 0.1.0
 
-### Bug Fixes
-
-* **opacity:** Opacity wasn't being exported to the SCSS file
-
 ### Features
 
 * **gulp:** Refactored gulp to process assets for publishing to NPM.
--- a/src/assets/package/README.md
+++ b/src/assets/package/README.md
@@ -6,7 +6,7 @@ A modular design system for Mozilla websites.
 
 Install package with NPM and add it to your dependencies:
 
-`npm install --save mozilla-protocol`
+`npm install --save @mozilla-protocol/core`
 
 ## Information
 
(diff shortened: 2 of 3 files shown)


michaelmccombie:
@alexgibson there are a couple of things we are going to have to discuss before we can get our main NPM package set up. The current package is name "protocol" which has already been taken. We wanted to publish everything under "@mozilla-protocol" so the name field would be "@mozilla-protocol/core" but in order to do so we need to upgrade our team to the paid tier that allows for private modules. 

For the time being, I published a package called [protocol-test](https://www.npmjs.com/package/protocol-test) which we can unpublish later once we have a solution.

alexgibson:
@michaelmccombie ok thanks, yes let's keep this open until we have the naming thing sorted.

I haven't used scoped packages before, but reading this I don't think it requires packages to be private (which we wouldn't want anyway) see: https://docs.npmjs.com/getting-started/scoped-packages#publishing-a-scoped-package

"By default, scoped packages are private. To publish private modules, you need to be a paid private modules user.

Public scoped modules are free and don't require a paid subscription. To publish a public scoped module, set the access option when publishing it. This option will remain set for all subsequent publishes."

```
npm publish --access=public
```

alexgibson:
We should make sure we update both the `package.json` and the [readme](https://github.com/mozilla/protocol/blob/master/src/assets/package/README.md) to match whatever the final package name will be.

michaelmccombie:
@alexgibson adding "--access=public" seemed to do the trick! I tested it with our assets design-tokens https://www.npmjs.com/package/@mozilla-protocol/assets 

Thank you for looking in to everything. I'll update this branch real quick

PR #47, opened 2018-04-19
Add general link styles

craigcook:
This updates/adds general link styling. I'm still using the old `_colors.scss` file at the moment since Protocol isn't yet consuming the proper tokens. I think some aspects may still be in flux so I've marked it as "inprogress" pending final decisions on color and style.

Meanwhile it can be previewed online and we'll finalize things before merging.

https://demo1--mozilla-protocol.netlify.com/patterns/atoms/links.html


craigcook:
Just submitted https://github.com/mozilla/protocol-design-tokens/pull/19 to update link color tokens, so we should do that so I can update this PR and get all the colors defined in the proper place.

PR #48, opened 2018-04-20
Import colors npm package

michaelmccombie:


src/data/colors.json
```diff
@@ -14,71 +14,71 @@
             "color": "Neon Blue",
```
craigcook: We should either change these names to match the token names or change the token names to match these display names. Just for consistency. 

The names came from the brand guide at https://mozilla.design/visualelements/ but we can deviate from that if we like the simpler names. If that sparks a larger conversation around color palette and naming it doesn't need to block this PR, I know the palette is going to keep evolving.

michaelmccombie: Yeah, some of those are difficult to work with "Dark Purple" for instance could lead to a variable called $color-darkpurple-dark which sounded odd to me, so I proposed changing the names in our Sketch library

image

michaelmccombie: protocol/includes/_colors.scss is just temporary until we can remove documentation site specific colors.

I want to refactor the "netlify.com/fundamentals/color.html" page to pull in data from the tokens package so we can replace the "src/data/colors.json" file. I'll set up an issue so we can tackle that in a new branch.

craigcook: Alright, not worth worrying about right now then.

PR #51, opened 2018-04-20
Include image assets in Protocol build task (Fixes #50)

### LINKED ISSUE #50: 404 image references when building from source via NPM
When running `./manage.py collectstatic` in bedrock the build fails with the following error:

```
Post-processing 'protocol/css/protocol.css' failed!
ValueError: The file 'img/cta-arrow.svg' could not be found
```

This seems to be because of a reference to a static image that is only part of the static demo site:

https://github.com/mozilla/protocol/blob/master/src/assets/sass/protocol/components/_card.scss#L93

We should move this out of `/static/` and into `/src/`, so we can bundle the required images via NPM. We should check for any other images also.

alexgibson:
- Separates Protocol image assets from docs site assets in the build process.
- Adds a global `$image-path` variable with a default value of  '/node_modules/@mozilla-protocol/core/protocol/protocol/img' (which will be used for the published module).
- Overrides this value in the docs with `'/assets/protocol/protocol/img'` (we can override in bedrock this way too).


alexgibson:
After making this change manually to test, I can confirm that `collectstatic` compiles OK in bedrock.

alexgibson:
Ok, I updated this PR to use a global `$image-path` variable which we can override as needed. The more I think this problem through, the better convinced I am that this is the most flexible solution. We may need to follow suit with the fonts here too.

alexgibson:
I'm going to close this PR as it makes sense to tacke the image and fonts both together, since they will share some of the same code.

PR #53, opened 2018-04-23
Allow for custom media paths in SCSS (Fixes #49, #50)

### LINKED ISSUE #49: Fonts are 404 once consumed by other sites
The Protocol `@font-face` declarations reference fonts located in `/static/`, which is a folder not copied across to NPM as it contains assets for the static docs site. This means when we pull Protocol into bedrock, the fonts are currently 404 as they reference a folder which doesn't exist. 

https://github.com/mozilla/protocol/blob/master/src/assets/sass/protocol/includes/fonts/_zilla-slab.scss#L10-L11

There are a couple of ways we could fix this:

- Copy the fonts over in our build script and switch to using relative paths. This is tidy, but may have the following issues:
  - It makes the NPM package very large as we need to bundle the fonts.
  - Sites may want to host fonts separately to their CSS & JS (e.g. a CDN, or Google Fonts).
  - Unless we're careful, we could end up loading fonts twice (e.g. global nav & footer still use Pebbles).
- The other option I can think of is not to include the `@font-face` declarations directly inside Protocol. Instead, we could have a separate file that sites could edit individually, pointing to where they host their fonts. This is more messy, but possibly more flexible as sites can edit the location.

Thoughts? /cc @craigcook @michaelmccombie

alexgibson:
- Includes Protocol img and font assets in the build.
- Adds global `$font-path` and `$image-path` variables with a default value of  `'/node_modules/@mozilla-protocol/core/protocol/protocol/{img|fonts}'` (which will be used for the published module).
- Overrides values in the docs with `'/assets/protocol/protocol/{img|fonts}'` (we can override in bedrock this way too).


alexgibson:
@craigcook @michaelmccombie not sure which of you has the most free time, but it would be good to try and get this reviewed/bikeshedded/merged as it's blocking getting Protocol into bedrock. Let me know if you have any questions, or would like to talk this through.

PR #86, opened 2018-05-02
Add article and sidebar menu organisms and main/sidebar template

craigcook:
This introduces a bunch of new bits and slightly updates a few existing bits. It looks bigger than it is, but it does have lots of small changes going on.

- New "article" organism.
- New "sidebar menu" organism.
- New "main with sidebar" template.
- New article demo page.
- New main-with-sidebar template demo page.
- New bi-directional mixin (swiped from MDN).
- New grid mixins that could maybe be reduced to just tokens for the widths.
- Updated unordered list styling.
- Renamed table styling.
- Adds tokens for media queries, content sizes, and spacing. These will eventually move to the tokens package but are included here for the moment.

Staged at https://demo2--mozilla-protocol.netlify.com/


src/assets/js/protocol/protocol-sidemenu.js
```diff
@@ -0,0 +1,20 @@
+(function() {
+    'use strict';
+
+    var $doc = document.documentElement;
+    var menu = document.querySelector('.mzp-c-sidemenu');
+    var menuToggle = document.querySelector('.mzp-js-toggle');
+
+    // Add class to reflect javascript availability for CSS
+    $doc.className = $doc.className.replace(/\bno-js\b/, 'js');
```
craigcook: This is may be redundant here and would collide with any other scripts doing the same thing. We should consider some kind of minimal global JS as part of protocol-core, even if all it does is set this class.

OR we can totally rethink it and not rely on this root-level script-is-working indicator. Each component can make its own check and set a class for a style hook for that component, but it sure is handy to have one hook in one place.

src/assets/sass/demos/article.scss
```diff
@@ -0,0 +1,567 @@
...
+        '../protocol/protocol';
+
+// Protocol components
+@import
+    '../protocol/templates/main-with-sidebar',
+    '../protocol/components/article',
+    '../protocol/components/sidebar-menu';
+
+
+// Article links... is this a component?
```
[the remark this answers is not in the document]
craigcook: This was a holdover from the security page this was originally based on, and on that page at least this set of links is a one-off thing. We'll have to figure out what to do about that when we're actually reskinning the security pages in bedrock, but I think for now I can just delete this bit here since we're not using it in the demo.

src/assets/sass/demos/article.scss
```diff
@@ -0,0 +1,567 @@
...
+        width: $width-desktop;
+    }
+
+    @media #{$mq-desktop-wide} {
+        width: $width-desktop-wide;
+    }
+}
+
+// Global nav
+.moz-global-nav {
```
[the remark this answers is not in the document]
craigcook: This was done as a test but you're right, we probably shouldn't keep it around when this demo is "officially" added. I'll probably keep a backup of the template locally but we can remove the header and footer from this PR.

src/assets/sass/docs/protocol.scss
```diff
@@ -6,3 +6,10 @@ $font-path:  '/assets/protocol/protocol/fonts';
 $image-path: '/assets/protocol/protocol/img';
 
 @import '../protocol/protocol.scss';
+
+// Import all the templates and components.
+// We need the full monolith for the doc site; you wouldn't import everything here in practice.
+@import
+    '../protocol/templates/main-with-sidebar',
```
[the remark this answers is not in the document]
craigcook: Yeah, this is something we've never really bikeshedded but probably should. I'm not sure where I got this multiline thing, I think when we first switched to Sass I followed someone else's style guide that encouraged it. But there seems to be no actual advantage to it and we should be more consistent with our style in other languages.

So new rule: no multiline @imports in Sass/SCSS. I'll update.

src/assets/sass/protocol/base/elements/_lists.scss
```diff
@@ -7,43 +7,53 @@
...
-    list-style: disc;
-    margin-left: 30px;
+ul.mzp-u-list-styled {
+    margin-left: $layout-sm;
 
     li {
         margin-bottom: .25em;
+
+        &:before {
+            content: '•';
```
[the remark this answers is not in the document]
craigcook: Yeah, just for size and spacing. We can talk to Vince about it but this was what bulleted lists looked like in the mockup so I tried to match it. I'm happy with browser defaults if the designers are.

src/assets/sass/protocol/components/_article.scss
```diff
@@ -0,0 +1,42 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import '../includes/lib';
+
+.mzp-c-article {
+    max-width: $content-md;
+
+    // Step down the heading levels
```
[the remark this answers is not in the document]
craigcook: A revision of the type scale is in the works.

src/assets/js/protocol/protocol-sidemenu.js
craigcook: I added a protocol-base.js that so far does nothing but this JS detection class but could possibly include other essentials down the road.

alexgibson:
One minor thing:

- The [concatJS](https://github.com/mozilla/protocol/blob/master/gulp/config.js#L61) gulp task is joining every JS file in `/js/protocol/` together, but we probably don't want `protocol-base.js` included in that, as it should be separate and in the `<head>`. Perhaps we want to exclude this file from concatenation?

src/assets/js/protocol/protocol-sidemenu.js
```diff
@@ -0,0 +1,20 @@
+(function() {
+    'use strict';
+
+    var $doc = document.documentElement;
+    var menu = document.querySelector('.mzp-c-sidemenu');
+    var menuToggle = document.querySelector('.mzp-js-toggle');
+
+    // Add class to reflect javascript availability for CSS
```
alexgibson: Should we remove this from the sidemenu JS now that we have a separate base JS file?

alexgibson:
This doesn't need to be a blocker, but currently we aren't compiling an CSS files for both the article and side menu. It's up to the consuming site to compile the sass (which might even be fine).

I think we probably need to decide if we even need to publish compiled CSS or contatenated / minified JS files (I'd imagine most sites would want to do this for themselves)

craigcook:
> This doesn't need to be a blocker, but currently we aren't compiling any CSS files for either the article or side menu, and it's up to the consuming site to compile the sass. This might be ok if it's what we decide we want to do, but it's currently not possible to include article.css (for example) in a page without sites having their own build step.

Yeah I've been thinking about the best way to publish components without creating one huge monolith. Dozens of individual CSS files isn't any better than one big one (probably worse). For bedrock we'll do our own compilation anyway and just @import the components we need per page, but longer term if we want to offer pre-compiled CSS we'll probably need to make a tool to pick your components, like https://foundation.zurb.com/sites/download.html/#customizeFoundation

In the short term we'll still need to get these components published and consumable as SCSS. Taking more inspiration from what Spark is doing, I think we'll need to make a separate package for protocol-components.

alexgibson:
> Yeah I've been thinking about the best way to publish components without creating one huge monolith. Dozens of individual CSS files isn't any better than one big one (probably worse). For bedrock we'll do our own compilation anyway and just @import the components we need per page, but longer term if we want to offer pre-compiled CSS we'll probably need to make a tool to pick your components, like https://foundation.zurb.com/sites/download.html/#customizeFoundation

> In the short term we'll still need to get these components published and consumable as SCSS. Taking more inspiration from what Spark is doing, I think we'll need to make a separate package for protocol-components.

In an http2 world, I think making individual CSS files is probably the better solution honestly. Sites will move away from bundling, an monolith CSS/JS files will be considered an anit-pattern. 

We could go all the way to publish a separate package a-la `protocol-components`, if we provided separate CSS files anyway, there's no real reason we couldn't just use one package still, and people could pick & choose what they use. Just my 2 cents

src/patterns/organisms/sidebar-menu/sidebar-menu.hbs
```diff
@@ -0,0 +1,39 @@
...
+  The menu summary reflects the current page, shown when the menu is collapsed in small viewports. The menu label should be the title of the site section.
+tips: |
+  - Indicate the current page with the `mzp-is-current` class.
+labels:
+    - prototype
+    - inprogress
+---
+
+<nav class="mzp-c-sidemenu">
+  <section class="mzp-c-sidemenu-summary mzp-js-toggle" aria-role="button" aria-controls="">
```
alexgibson: I think the value for `aria-controls` should be the ID of the menu?

src/assets/js/protocol/protocol-sidemenu.js
```diff
@@ -0,0 +1,30 @@
...
+    menuMain.setAttribute('aria-expanded', 'false');
+
+
+    // Toggle the sidebar menu
+    menuToggle.addEventListener('click', function(e) {
+        e.preventDefault();
+        menu.classList.toggle('mzp-is-active');
+
+        menuExpanded = menuMain.getAttribute('aria-expanded');
+        if (menuExpanded == 'true') {
```
alexgibson: You could make this logic a little simpler with something like:

```javascript
 var menuExpanded = menuMain.getAttribute('aria-expanded') == 'true' ? 'false' : 'true';
```

src/assets/js/protocol/protocol-sidemenu.js
```diff
@@ -0,0 +1,30 @@
+(function() {
+    'use strict';
+
+    var menu = document.querySelector('.mzp-c-sidemenu');
+    var menuMain = document.querySelector('.mzp-c-sidemenu-main');
+    var menuToggle = document.querySelector('.mzp-js-toggle');
+    var menuExpanded;
```
alexgibson: I don't think this variable needs to be this scope for any reason, you can probably just declare it inside the `click` even listener?

src/patterns/organisms/sidebar-menu/sidebar-menu.hbs
alexgibson: Should both this and `aria-role="button"` be set in the JS along with where you set `tabindex`? They probably don't apply when JS is disabled?

src/assets/js/protocol/protocol-sidemenu.js
```diff
@@ -0,0 +1,30 @@
...
+        menu.classList.toggle('mzp-is-active');
+
+        menuExpanded = menuMain.getAttribute('aria-expanded');
+        if (menuExpanded == 'true') {
+            menuExpanded = 'false';
+        } else {
+            menuExpanded = 'true';
+        }
+        menuMain.setAttribute('aria-expanded', menuExpanded);
+        alert(menuExpanded);
```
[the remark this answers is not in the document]
craigcook: Doh.

craigcook:
🍾

PR #112, opened 2018-05-15
Add newsletter form and basic button

craigcook:
This adds a newsletter subscription form molecule as it appears on the home page. We'll probably come up with a handful of variants in due time. It also adds a basic button style that we probably need to expand upon, but the form needed a button so we might as well make it an atom.

Marking this WIP initially since there should be some discussion around the button style as well as how to handle the signup form functionality. 

Right now this uses a slight reformulation of the basket-client.js from https://github.com/mozilla/basket-example/ but upon further thought I think we should keep the Protocol component minimal and refer people to that external repo for the subscription functionality. I'm working on that next so updates to this PR are forthcoming.

https://demo2--mozilla-protocol.netlify.com/patterns/organisms/newsletter.html
https://demo2--mozilla-protocol.netlify.com/demos/newsletter.html


src/assets/sass/protocol/components/_newsletter-form.scss
```diff
@@ -0,0 +1,111 @@
...
+    input[type='email'],
+    select {
+        width: 100%;
+    }
+
+    .mzp-c-form-submit {
+        text-align: center;
+    }
+
+    .mzp-c-button {
```
alexgibson: Would this be worth having as a button style atom itself, as opposed to something specific to the newsletter form?

src/patterns/organisms/newsletter/newsletter.hbs
```diff
@@ -0,0 +1,86 @@
...
+            <label for="format-text" class="mzp-u-inline">
+              <input type="radio" id="format-text" name="fmt" value="T"> Text
+            </label>
+          </p>
+        </fieldset>
+
+        <p>
+          <label for="privacy" class="mzp-u-inline">
+            <input type="checkbox" id="privacy" name="privacy" required aria-required="true">
+            I’m okay with Mozilla handling my info as explained in <a href="#">this Privacy Notice</a>
```
alexgibson: Should we link to the actual privacy policy in the example?

src/patterns/organisms/newsletter/newsletter.hbs
```diff
@@ -0,0 +1,86 @@
...
+      </div>
+
+      <p class="mzp-c-form-submit">
+        <button type="submit" class="mzp-c-button" id="newsletter-submit">Sign up now</button>
+        <span class="mzp-c-fieldnote">We will only send you Mozilla-related information.</span>
+      </p>
+    </fieldset>
+  </form>
+
+  <div id="newsletter-thanks" class="mzp-c-newsletter-thanks">
```
alexgibson: Is there any way to see these styles in the example without the form being functional?

alexgibson:
Is `src/static/img/placeholder.png` used anywhere?

src/assets/js/protocol/protocol-newsletter.js
```diff
@@ -0,0 +1,41 @@
...
+
+    // Expand email form on input focus or submit if details aren't visible
+    function initEmailForm() {
+        var newsletterForm = document.getElementById('newsletter-form');
+        var submitButton = document.getElementById('newsletter-submit');
+        var formDetails = document.getElementById('newsletter-details');
+        var formControls = document.querySelectorAll('.mzp-c-newsletter-form input, .mzp-c-newsletter-form select');
+
+        function emailFormShowDetails() {
+            if (window.getComputedStyle(formDetails).display === 'none') {
```
alexgibson: Rather than use these`getComputedStyle` calls, it may be slightly more performant to use a boolean variable to determine if the form has expanded? e.g. 

```javascript
if (!formExpanded) {
    formDetails.style.display = 'block';
    formExpanded = true;
}
```

alexgibson:
RTL styles are a little busted https://screenshots.firefox.com/qShhAExCQZXpa4gN/localhost

alexgibson:
`src/newsletter-image.png` seems like it may be a leftover?

src/assets/sass/protocol/components/_newsletter-form.scss
craigcook: Quite possibly, but I think Vince/Jennifer are still working on some button styles so I don't know if this will stick.

src/patterns/organisms/newsletter/newsletter.hbs
craigcook: I've set up the demo page to show different states of the form (normal, errors, success).

src/assets/js/protocol/protocol-newsletter.js
```diff
@@ -0,0 +1,47 @@
...
+        function emailFormShowDetails() {
+            if (!formExpanded) {
+                formDetails.style.display = 'block';
+                formExpanded = true;
+            }
+        }
+
+        for (var i = 0; i < formControls.length; i++) {
+            formControls[i].addEventListener('focus', function() {
+                emailFormShowDetails();
```
alexgibson: Do we need to add `if (!formExpanded) { }` here?

src/assets/js/protocol/protocol-newsletter.js
```diff
@@ -0,0 +1,47 @@
...
+
+    // Expand email form on input focus or submit if details aren't visible
+    function initEmailForm() {
+        var newsletterForm = document.getElementById('newsletter-form');
+        var submitButton = document.getElementById('newsletter-submit');
+        var formDetails = document.getElementById('newsletter-details');
+        var formControls = document.querySelectorAll('.mzp-c-newsletter-form input, .mzp-c-newsletter-form select');
+        var formExpanded;
+
+        if (window.getComputedStyle(formDetails).display === 'none') {
```
alexgibson: If this is condition falsy then the value for `formExpanded` will be `null` instead of `true`. Maybe something like this could work?

```javascript
var formExpanded = window.getComputedStyle(formDetails).display === 'none' ? false : true;
```

src/pages/demos/newsletter.hbs
```diff
@@ -0,0 +1,104 @@
...
+<div class="mzp-l-content">
+
+  {{#embed "patterns.organisms.newsletter.newsletter"}}{{/embed}}
+
+  <hr>
+
+  <p>With errors.</p>
+
+  <div class="mzp-qa-newsletter-errors">
+<aside class="mzp-c-newsletter">
```
alexgibson: Nit: indentation

src/assets/js/protocol/protocol-newsletter.js
craigcook: The `emailFormShowDetails` function already checks for it, but we need to check if the form is expanded on the submit button just for the sake of `e.preventDefault`. I'm sure there's a more elegant way to do this... I'm not good at JavaScript.

PR #113, opened 2018-05-15
Update home page

craigcook:
Adds some stub content to the home page so it's a tad less empty. Also updates the favicon and removes the blank Spacing page (that will come back once we can spend a little more time on that).


thejeffreytaylor:
I added 1 word.

PR #118, opened 2018-05-18
Add modal organism (Fixes #78)

### LINKED ISSUE #78: [home page] Create modal organism
- Home page video cards open modal on click

alexgibson:
- Adds modal organism (replicating style seen on the [new mozilla.org home page](https://www.mozilla.org/en-US/?v=a))
- Adds a `light-links` mixin, for better link contrast on dark backgrounds.

Note: I had originally hoped to use `<dialog>` and a polyfill, but I discovered some a11y shortcomings in the polyfill that I don't think we can lose to justify switching yet 😢. Maybe once more browsers have native support. For now, this is a direct port of the bedrock modal JS (just replacing the jQuery for native JS).

~We still need to bikeshed the styles here (including things like the close icon, which needs to be moved to the design assets repo once we decide what to use).~

Demo: https://demo1--mozilla-protocol.netlify.com/patterns/organisms/modal.html


src/assets/img/icons/ui/close-lg.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 42 42"><style>.st0{fill:none;stroke:#fff;stroke-miterlimit:10}</style><path class="st0" d="M0 0l42 42M0 42L42 0"/></svg>
```
alexgibson: Once we decide on an icon to use, this need to be moved to the design assets repo.

src/assets/sass/protocol/components/_modal.scss
```diff
@@ -0,0 +1,133 @@
...
+
+                a:link,
+                a:visited {
+                    color: $color-secondary;
+                }
+
+                a:hover,
+                a:focus,
+                a:active {
+                    color: darken($color-secondary, 5%);
```
alexgibson: Should this be a new link style atom?

src/assets/sass/protocol/components/_modal.scss
```diff
@@ -0,0 +1,130 @@
...
+
+    a:link,
+    a:visited {
+        color: $color-secondary;
+    }
+
+    a:hover,
+    a:focus,
+    a:active {
+        color: darken($color-secondary, 5%);
```
alexgibson: Should this link style be a new atom?

craigcook: I've been thinking we probably need to standardize a style for links on dark backgrounds and make a mixin for it, something like `@include light-links;`

alexgibson: Added a `light-links` mixin 👍

alexgibson:
Ok, looks like we already have a close icon in the design-assets repo, so I updated the modal here to reuse that instead of introduce a new icon. It also has slightly thicker paths, so contrast is improved.

src/patterns/organisms/modal/modal.hbs
```diff
@@ -0,0 +1,60 @@
...
+
+    var content = document.querySelector('.mzp-u-modal-content');
+    var button = document.querySelector('.mzp-c-button');
+
+    button.addEventListener('click', function(e) {
+        e.preventDefault();
+
+        Mozilla.Modal.createModal(e.target, content, {
+          title: 'Example headline with 35 characters',
+          closeText: 'Close modal',
```
craigcook: We probably need some better documentation around this, but we can at least comment that this is a demo and ideally one shouldn't have translatable strings in JS (unless you have a system for translating JS files or aren't localizing anyway). The title could be a heading in the DOM, closeText could be a data attribute, etc.

src/assets/sass/protocol/components/_modal.scss
```diff
@@ -0,0 +1,121 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import '../includes/lib';
+@import '../includes/animations';
+
+html.noscroll {
```
craigcook: `mzp-is-noscroll`?

src/assets/js/protocol/protocol-modal.js
```diff
@@ -1 +1,159 @@
...
+            '      <header><h2>' + title + '</h2></header>' +
+            '      <div class="mzp-c-modal-close">' +
+            '        <button type="button" class="mzp-c-modal-button-close">' + closeText + '</button>' +
+            '      </div>' +
+            '    </div>' +
+            '  </div>' +
+            '</div>';
+
+        if ((options && !options.allowScroll) || isSmallViewport) {
+            html.classList.add('noscroll');
```
craigcook: Should this class follow the naming convention for states? So it would be `mzp-is-noscroll`

src/patterns/organisms/modal/modal.hbs
```diff
@@ -0,0 +1,60 @@
...
+          onCreate: function() {
+              console.log('Modal opened');
```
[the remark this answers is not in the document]
craigcook: On second thought the notes already explain what onCreate and onDestroy do and this whole thing is clearly just an example already, maybe no need to drive that point home any further.

src/patterns/organisms/modal/modal.hbs
alexgibson: I think it may be best (for now at least) to try and shy away from too much hand-holding when it comes to l10n, as different sites all have different methods of doing things. It's not directly related to the component itself, so to avoid over-complicating the instructions here, I think a generic l10n section may be a better time to explain this type of thing?

alexgibson:
Thanks @craigcook - I updated the `noscroll` class, and also added a basic focus outline for the close button (better than the miniscule scale effect).

src/patterns/organisms/modal/modal.hbs
craigcook: Yeah, I'm probably overly paranoid about people following this example when hopefully it's clear that it's JUST an example. We're working on general l10n guidelines and "no strings in JS" is a good rule.

PR #123, opened 2018-05-23
Use handlebars templates for card examples (Fixes #108)

### LINKED ISSUE #108: Create macro for card component markup examples
The card componenet examples currently repeat a bunch of markup for each demo example, but we could build a macro to include in templates to make updates easier and more DRY.

alexgibson:
Makes card component examples more DRY using handlebars templates.

### DIFF
--- a/src/pages/demos/card-layout.hbs
+++ b/src/pages/demos/card-layout.hbs
@@ -24,70 +24,11 @@ styles:
   </header>
 
   <div class="mzp-l-card-hero">
-    <section class="mzp-c-card mzp-c-card-large mzp-has-aspect-16-9">
-      <a class="mzp-c-card-block-link" href="#">
-        <div class="mzp-c-card-media-wrapper">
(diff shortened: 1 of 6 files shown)


alexgibson:
The current names are what we came up with in Portland together with @vincejoyiv. I’m happy to bikeshed them further, or change to any names we think work better. But I think doing that is likely out of scope for this particular PR?

craigcook:
Correct, nothing to do with this PR. It only became clear to me on seeing `{{#embed "patterns.molecules.card.small-card" class="mzp-c-card-large mzp-has-aspect-16-9"}}{{/embed}}` in the large-card template how weird it seemed the "small" was our baseline default. We can discuss later.

alexgibson:
Yup, it does read a little confusing. This is only really in the handlebars template though, which is only really for the docs site. The actual css class names used for the component don’t use `card-small` as a base class name. Still, we could certainly rename the cards in the docs to be better reflective of this.

PR #135, opened 2018-06-04
Media content should occupy 100% modal width (Fixes #132)

### LINKED ISSUE #132: Modal videos and images should have an explicit width of 100%
Currently if a video or image is displayed within a modal component, we set [max-width: 100%;](https://github.com/mozilla/protocol/blob/master/src/assets/sass/protocol/components/_modal.scss#L113). This looks a little odd however if the media is smaller than the maximum modal width.  We should probably just set `width: 100%`, and make sure to put some modal guidelines in the docs for preferred media sizes.

alexgibson:


craigcook:
This is good if the media really should fill the entire modal, but would cause problems for other kinds of content. I'm thinking of the modals on the leadership page that include a headshot; those images would be blown up.

Should we do a special "media" variant with a modifier class? Then the base modal can be content-agnostic.

alexgibson:
> Should we do a special "media" variant with a modifier class? Then the base modal can be content-agnostic.

This is a good suggestion, I'll make it so 👍

alexgibson:
@craigcook thanks for the suggestion, updated.

Issue #109, opened 2018-05-14
Refactor Protocol to consume tokens


michaelmccombie:
We need to refactor protocol to consume all of our design tokens and to work with any new naming conventions.

alexgibson:
It would be nice to try and get this one done soon, as I'm trying to reference CSS variables in the mozorg home page for colors that don't exist in the older tokens repo 👍

alexgibson:
Blocked by https://github.com/mozilla/protocol-tokens/issues/9

alexgibson:
Blocked by https://github.com/mozilla/protocol-tokens/pull/14

alexgibson:
Blocked by https://github.com/mozilla/protocol-tokens/issues/16

alexgibson:
\o/

PR #149, opened 2018-06-25
Add page footer organism (Fixes #97) 

### LINKED ISSUE #97: [home page] Port site footer

alexgibson:
This branch builds on #109, so I'll rebase once that is reviewed & merged.

- Add page footer organism
- Add language switcher molecule
- Add Karma & Jasmine for JS unit testing
- Run JS tests in Travis CI against pull requests.


alexgibson:
Ok, this is rebased and ready for review

alexgibson:
Rebased

src/patterns/organisms/footer/footer.hbs
```diff
@@ -0,0 +1,66 @@
...
+
+<footer class="mzp-c-footer mzp-has-lang-switcher">
+  <div class="mzp-l-content">
+    <nav class="mzp-c-footer-primary">
+      <div class="mzp-c-footer-primary-logo"><a href="https://www.mozilla.org/">Mozilla</a></div>
+      <section class="mzp-c-footer-section">
+        <h5>
+          <a href="https://www.mozilla.org/">Mozilla</a>
+        </h5>
+        <ul class="mzp-c-footer-links-mozilla">
```
craigcook: Are these lists (this one and the `mzp-c-footer-links-firefox` one below) distinctive enough to get specific component classes? It seems like they're both common lists and we could use a class on the parent section to distinguish that one is Mozilla and one is Firefox (and even that distinction may not be necessary).

src/patterns/organisms/footer/footer.hbs
```diff
@@ -0,0 +1,66 @@
...
+              <li><a class="twitter" href="https://twitter.com/firefox">Twitter<span> (@firefox)</span></a></li>
+              <li><a class="youtube" href="https://www.youtube.com/firefoxchannel">YouTube<span> (firefoxchannel)</span></a></li>
+            </ul>
+          </li>
+        </ul>
+      </section>
+    </nav>
+
+    <nav class="mzp-c-footer-secondary">
+      <div class="mzp-c-footer-links-small">
```
craigcook: Nit: this feels like a presentational name. Maybe something like `mzp-c-footer-legal` (since the div holds both the links and the license blurb).

src/assets/js/protocol/protocol-lang-switcher.js
```diff
@@ -0,0 +1,63 @@
...
+            window.location.href = destination;
+        }
+    };
+
+    /**
+     * Initialize footer lang switcher.
+     * @param {function} Custom callback for analytics.
+     */
+    LangSwitcher.init = function(callback) {
+        var language = document.getElementById('mzp-c-language-switcher-select');
```
craigcook: Since we're using an ID does that mean we can only allow one language switcher per page? If so we should mention that in the usage notes. Or if we want to allow multiples we'd need to rework this a bit. I could imagine someone wanting a language switcher at both the top and bottom of a page (the elements would still need unique IDs when the molecule is implemented but we could leave the values up to the implementor).

But I'm also fine with enforcing it as a singleton, which at least covers our immediate needs on mozorg. So let's just note that we currently only support one switcher per page and we can always revisit later (especially since the footer is likely getting overhauled soon anyway).

src/patterns/organisms/footer/footer.hbs
alexgibson: Neither are referenced in the CSS, so I think we're fine to remove these entirely 👍

src/assets/js/protocol/protocol-lang-switcher.js
alexgibson: I think it's pretty safe to assume that at one point, a site may want both a lang switcher in the header and footer of a page. My initial thought was let's wait to see if that requirement comes before changing the existing functionality that we've ported here from bedrock, but looking at the code I think this is a simple enough change to make. Let's do this now 👍

alexgibson:
Thanks for the review @craigcook!

I added a second commit with some review fixes.

I also noticed a regression in the billboard demo page that was introduced in https://github.com/mozilla/protocol/pull/166. I added a third commit which fixes that (changing the demo page to use a handlebars template to keep things better in sync). As this was related only to the docs site, I figured it was OK to include here as opposed to a separate issue.

alexgibson:
Oh, I also bumped Protocol to 2.0.0 here, so we can publish once this merges 🎉

PR #162, opened 2018-07-02
Add utility classes for background colors (Fixes #151)

### LINKED ISSUE #151: Utility Classes for Background Colors
Create utility classes (i.e. `.background-blue`) that apply a background color to a component using the color design tokens

alexgibson:
Fixes: https://github.com/mozilla/protocol/issues/151

This PR adds a bunch of common utility classes for background colors. 

While I understand the need for this type of utility class, these class names are largely presentational and something we try to avoid in general. Class names such as `mzp-u-bg-light` and `mzp-u-bg-dark` are fine, but `mzp-u-bg-color-teal-70` et al feel overly specific.

Issue #151 doesn't go into too much detail on which colors from our pallet we really need, so I wonder if we can come up with some sort of general theming classes/mixins as opposed to what we have here.

Interested in both @michaelmccombie and @craigcook's thoughts

### DIFF
--- a/src/assets/sass/protocol/includes/_colors.scss
+++ b/src/assets/sass/protocol/includes/_colors.scss
@@ -1,0 +1,275 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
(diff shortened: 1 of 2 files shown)


alexgibson:
Ok, we've discussed this on Slack and are going to limit the colors via some theming mixins instead.

PR #164, opened 2018-07-03
Update card layout demo title to "6 Card Layout"

alexgibson:
For reference: https://github.com/mozilla/bedrock/issues/5865#issuecomment-402071643

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Bug Fixes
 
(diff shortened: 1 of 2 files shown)


alexgibson:
After discussing this with MJ Kelly yesterday, I think it makes the most sense to change the title on the demo page from "3 Card Layout" to "6 Card Layout" to avoid confusion, since an even number of cards is what we suggest using.

PR #172, opened 2018-07-09
Add Call Out component (Fixes #153)

### LINKED ISSUE #153: Call Out Section
This section can include any combination of these four elements:
- Logo
- Headline
- Description
- CTA

The plan is for this section to be theme-able. The Mozilla theme, for instance, would use the Zilla-Slab font for its headlines.

image

alexgibson:
- Adds Call Out component and compact variation.

Demo:
- https://mozilla-protocol.netlify.com/patterns/organisms/call-out.html
- https://mozilla-protocol.netlify.com/demos/call-out.html
- https://mozilla-protocol.netlify.com/demos/call-out-compact.html

### DIFF
--- a/src/assets/sass/protocol/components/_call-out.scss
+++ b/src/assets/sass/protocol/components/_call-out.scss
@@ -1,0 +1,219 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import '../includes/lib';
+
+/* -------------------------------------------------------------------------- */
+// Call Out component
+
+.mzp-c-call-out {
+    background-color: $color-gray-20;
+    text-align: center;
+
+    &.mzp-t-dark {
+        background-color: $color-gray-90;
+        color: $color-white;
+
+        .mzp-c-call-out-desc {
+            color: $color-gray-30;
+        }
+    }
+
+    .mzp-c-call-out-title {
+        @include text-display-md;
+    }
+
+    .mzp-c-call-out-desc {
+        @include text-body-lg;
+        margin-bottom: 0;
+    }
+
+    .mzp-c-button-download {
+        margin: $margin-lg 0 0;
+    }
+
+    &.mzp-t-product-firefox ,
+    &.mzp-t-product-beta,
+    &.mzp-t-product-developer,
+    &.mzp-t-product-nightly,
+    &.mzp-t-product-focus {
+        .mzp-c-call-out-title {
+            @include background-size(64px 64px);
+            background-position: top center;
+            background-repeat: no-repeat;
+            padding: (64px + $padding-lg) 0 0 0;
+        }
+    }
+
+    &.mzp-t-product-firefox .mzp-c-call-out-title {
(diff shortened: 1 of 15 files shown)


src/assets/sass/protocol/components/_call-out.scss
```diff
@@ -0,0 +1,219 @@
...
+
+    .mzp-c-button-download {
+        margin: $margin-lg 0 0;
+    }
+
+    &.mzp-t-product-firefox ,
+    &.mzp-t-product-beta,
+    &.mzp-t-product-developer,
+    &.mzp-t-product-nightly,
+    &.mzp-t-product-focus {
```
michaelmccombie: Instead of writing out these class names it might make more sense to have one theme `.mzp-t-product-firefox` that acts as a generic firefox theme and if and if you wanted to theme something to a sub product, Firefox Focus for instance, we would add both the `.mzp-t-product-firefox` class and the `.focus` class to the component.

src/assets/sass/protocol/components/_call-out.scss
```diff
@@ -0,0 +1,219 @@
...
+            background-repeat: no-repeat;
+            padding: (64px + $padding-lg) 0 0 0;
+        }
+    }
+
+    &.mzp-t-product-firefox .mzp-c-call-out-title {
+        @include at2x('#{$image-path}/logos/firefox/firefox.png', 64px, 64px);
+    }
+
+    &.mzp-t-product-beta .mzp-c-call-out-title {
```
michaelmccombie: So this line would become something like: `&.mzp-t-product-firefox.beta .mzp-c-call-out-title {`

src/assets/sass/protocol/components/_call-out.scss
```diff
@@ -0,0 +1,219 @@
...
+        }
+
+        .mzp-c-call-out-title + .mzp-c-call-out-desc {
+            margin-top: $margin-sm;
+        }
+
+        .mzp-c-call-out-content {
+            @include bidi(((float, left, right),));
+            max-width: 624px;
+            width: calc(66% - #{$margin-lg});
```
michaelmccombie: We should try and be consistent with our fractions. I've seen percentages with decimals in other files (33.3%). My personal preference is to use calc(100% * 2/3) instead of 66% so we don't have to deal with fractions of pixels and layout issues

src/assets/sass/protocol/components/_call-out.scss
alexgibson: Good call 👍

src/assets/sass/protocol/components/_call-out.scss
alexgibson: I'll update to `66.6%` here, but like I mentioned in the other PR I don't think there's any real benefit in using `calc()` here

src/assets/sass/protocol/components/_call-out.scss
alexgibson: On second thought, I think this maybe requires a little more consideration. The one snag here is introducing a non-mzp prefixed class name, which is something we should discourage in general to avoid conflicts in third-party CSS. Adding too many `mzp` class names to a single component for theming also gets a little verbose.

In the future we may also want to add more products to this that don't fall into the Firefox line of products, so perhaps this is the more future proof method for now.

alexgibson:
Many thanks for the review @michaelmccombie 💯

Issue #155, opened 2018-06-27
Section with two columns


michaelmccombie:
**Desktop (1024px)**
<img width="573" alt="image" src="https://user-images.githubusercontent.com/5924575/41959048-ca602ea2-79a0-11e8-9df5-c84d05a60e8d.png">

**Mobile (480px)**
<img width="288" alt="image" src="https://user-images.githubusercontent.com/5924575/41959071-d4f4f190-79a0-11e8-9573-a78bb2f1bd48.png">

alexgibson:
It makes sense to bundle this with #154

alexgibson:
Closed via https://github.com/mozilla/protocol/commit/6e0bd14a134fd35b5063fe7501eafa8d3e89cf87

PR #173, opened 2018-07-10
Add Picto Card component (Fixes #154, #155)

### LINKED ISSUE #154: Section with three columns
**Variations**
- Text can be either center or left-aligned
- Background color should be interchangeable

**Desktop (1024px)**
image

**Mobile (480px)**
image

alexgibson:
Fixes #154, #155.

- Adds a new Picto Card component with compact variation (happy to choose a different name, but I thought this sounded better than something like "Icon Card").
- Updates card layout classes to support groups of 2 & 3 Picto Cards.

Demo:
- https://mozilla-protocol.netlify.com/patterns/molecules/card.html#picto-card
- https://mozilla-protocol.netlify.com/patterns/molecules/card.html#picto-card-compact
- https://mozilla-protocol.netlify.com/patterns/molecules/card.html

### DIFF
--- a/src/assets/sass/demos/card.scss
+++ b/src/assets/sass/demos/card.scss
@@ -8,3 +8,7 @@
 @import '../protocol/includes/lib';
 @import '../docs/protocol';
 @import '../docs/protocol-extra';
+
+.mzp-t-dark {
+    background-color: $color-gray-90;
+}
--- a/src/assets/sass/protocol/components/_card.scss
+++ b/src/assets/sass/protocol/components/_card.scss
@@ -166,3 +166,95 @@
     }
 }
 
+/* -------------------------------------------------------------------------- */
+// A small card with an icon and no link.
+
+.mzp-c-card-picto {
+    color: $color-black;
+    padding: $padding-xl 0;
+    text-align: center;
+
+    .mzp-c-card-picto-content {
+        padding-top: 56px + $padding-lg;
+        position: relative;
+
+        &:before {
+            background: $color-gray-20;
+            content: '';
+            height: 56px;
+            left: 50%;
+            margin-left: -28px;
+            position: absolute;
+            top: 0;
+            width: 56px;
+        }
+    }
+
+    .mzp-c-card-picto-title {
+        @include text-display-xs;
+    }
+
+    .mzp-c-card-picto-desc {
+        @include text-body-md;
+        color: $color-gray-70;
+        margin-bottom: 0;
+    }
+
+    // dark theme color styling
+    .mzp-t-dark & {
+        color: $color-white;
+
+        .mzp-c-card-picto-desc {
+            color: $color-gray-30;
+        }
+    }
+
+    @media #{$mq-md} {
+        padding: $padding-2xl 0;
+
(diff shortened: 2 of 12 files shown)


src/assets/sass/demos/card.scss
```diff
@@ -8,3 +8,7 @@
 @import '../protocol/includes/lib';
 @import '../docs/protocol';
 @import '../docs/protocol-extra';
+
+.mzp-t-dark {
```
michaelmccombie: Demos should probably use the "protosite" prefix or some other sort of prefix. Especially with classes like this one which will probably end up in the system at some point.

src/assets/sass/protocol/components/_card.scss
```diff
@@ -166,3 +166,95 @@
...
+
+        &:before {
+            background: $color-gray-20;
+            content: '';
+            height: 56px;
+            left: 50%;
+            margin-left: -28px;
+            position: absolute;
+            top: 0;
+            width: 56px;
```
michaelmccombie: Not a blocker. 56px is the right spec for the landing page icons and I'll make sure designers stick to that size for the next couple of projects.
But I could see a scenario where we have to support larger icons (similar to what we have on the old homepage https://www.mozilla.org/de/). So we might have to add "mzp-c-card-picto-small", "mzp-c-card-picto-medium", etc. classes at some point.

src/assets/sass/protocol/templates/_card-layout.scss
```diff
@@ -158,6 +158,23 @@
...
+/* -------------------------------------------------------------------------- */
+// 3 picto card layout
+// On small screens cards stack vertically and span full container width.
+// On medium screens cards span 1 third width.
+.mzp-l-card-third {
+    @media #{$mq-md} {
+        @include clearfix;
+
+        .mzp-c-card-picto {
+            width: 33%;
```
michaelmccombie: It's a minor thing but we should try and use `calc(100% / 3)` instead of 33% to keep widths from becoming fractions of pixels and keep the layout consistent.

At the moment we are using the Bedrock's legacy breakpoint/ max-width so we won't be able to completely avoid fractions of pixels, but I'm hoping to adjust those values when we have almost everything running protocol.

There are also a few places where we are using 33.3% (_card-layout.scss).

alexgibson: I believe browsers generally round down decimal CSS values when it comes to rendering, so I think it's probably no different to use `33.3%` without `calc()` here, rather than  ending up with a value like `33.3333333%`.

src/assets/sass/demos/card.scss
alexgibson: Hmm, I'm not sure here - the demo is intended as an instructional example of how to use the component, so I don't think it's something related to the docs site itself.

src/assets/sass/protocol/components/_card.scss
alexgibson: Once we have a need for larger icons we can update 👍

alexgibson:
Thanks again @michaelmccombie 👍

Issue #174, opened 2018-07-16
Error compiling published source files due to missing tokens


alexgibson:
When updating bedrock to import `@mozilla-protocol/core` v2.0.0, Sass compilation fails with the following error:

```
assets_1 | messageFormatted: node_modules/@mozilla-protocol/core/protocol/css/includes/_lib.scss
assets_1 | Error: File to import not found or unreadable: tokens/dist/index.scss.
assets_1 | on line 6 of node_modules/@mozilla-protocol/core/protocol/css/includes/_lib.scss
assets_1 | from line 8 of static_build/css/mozorg/home/home-2018.scss
assets_1 | >> @import 'tokens/dist/index.scss';
```

Looking at the `/dist/` directory in Protocol, it looks like our token source `.scss` files aren't being included correctly in the build (even though the compiled CSS builds fine, since Protocol looks in `node_modules` for the dependency).

PR #175, opened 2018-07-16
Include @mozilla-protocol/tokens source files in build (Fixes #172)

alexgibson:
Fixes https://github.com/mozilla/protocol/issues/174

### DIFF
--- a/gulp/config.js
+++ b/gulp/config.js
@@ -96,8 +96,8 @@ module.exports = {
             dest: `${dest}/docs/css/`
         },
         includes: {
-            src: `./node_modules/@mozilla-protocol/*/*.scss`,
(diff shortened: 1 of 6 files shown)


alexgibson:
Confirmed this fixes the issue locally in bedrock

gulp/config.js
```diff
@@ -96,8 +96,8 @@ module.exports = {
             dest: `${dest}/docs/css/`
         },
         includes: {
-            src: `./node_modules/@mozilla-protocol/*/*.scss`,
```
alexgibson: This selector meant we we're only looking one level deep :/

craigcook: what a difference an asterisk makes...

PR #183, opened 2018-07-24
[fix #157, fix #158] Add hero section organism

### LINKED ISSUE #157: Hero Section - Centered Text
**Variants**
- The background color is interchangeable 
- A background image can be applied

**Notes**
- It would nice (definitely not a requirement) if we could apply a dynamic bezier curve animation similar to this example.
https://www.joshwcomeau.com/posts/dynamic-bezier-curves

**Desktop (1024px)**
image
image

**Mobile (480px)
image

### LINKED ISSUE #158: Hero Section - Mozilla Theme
Similiar to the other hero section components.
- Zilla-Slab as the headline font. 
- Doesn't include a logo. 
- Includes an image which can be either left or right aligned.

Desktop (1024px)
image

Mobile (1024px)
image

craigcook:
This adds a hero section organism with several variations. 

It uses an SVG mask to draw the bottom shape in a way that makes it easy to change the color with CSS alone. This could be done with inline SVG as well but it's a presentational image so CSS and a mask felt more semantically correct, even if it's a bit more bleeding edge support-wise ([check caniuse](https://caniuse.com/#search=mask); it's a plain straight line in unsupporting browsers). It could just as easily be a background image but then we'd lose the easy color changing (would need to swap images). That's certainly the simpler option if we don't expect the color to change often but I thought a mask was worth a try.

See https://github.com/mozilla/protocol-assets/pull/10 for adding the SVG shape to protocol-assets. That PR kinda blocks this one but not really.

This also adds an inverted button style for the dark theme.

Organism: https://demo1--mozilla-protocol.netlify.com/patterns/organisms/hero.html
Demo: https://demo1--mozilla-protocol.netlify.com/demos/hero.html


src/assets/sass/protocol/components/_hero.scss
```diff
@@ -0,0 +1,224 @@
...
+
+@supports (mask-image: url('mask.svg')) {
+    .mzp-c-hero.mzp-t-firefox {
+        &:after {
+            background-color: $color-gray-20;
+            bottom: 0;
+            content: '';
+            display: block;
+            height: 25%;
+            mask: url('#{$image-path}/masks/hero/hero-curve.svg') center bottom no-repeat;
```
stephaniehobson: 🔥

craigcook:
Now that https://github.com/mozilla/protocol-assets/pull/10 has merged this is ready for review

craigcook:
Hold on, need to work a bit more on the image positioning...

craigcook:
OK, ready now.

src/assets/sass/protocol/components/_hero.scss
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    .mzp-t-product-focus & {
+        @include at2x('#{$image-path}/logos/firefox/focus.png', 80px, 80px);
+    }
+}
+
+.mzp-c-hero-desc {
+    @include text-body-lg;
+    margin-bottom: $spacing-md;
```
alexgibson: Are we abandoning all `$padding` and `$margin` token values? I know they are the same basic value, but if we're going to use something in protocol (or not) I think we should probably make a decision on it and use one thing uniformally. This could be an issue to bikeshed :)

src/pages/demos/hero.hbs
```diff
@@ -0,0 +1,173 @@
...
+      </div>
+
+      <p class="mzp-c-hero-cta">
+        <a class="mzp-c-button mzp-c-button-primary" href="#">Call to action</a>
+      </p>
+    {{/content}}
+
+  {{#content "hero-image"}}
+  <div class="mzp-c-hero-image">
+    <img src="https://placeimg.com/1200/675/nature" alt="">
```
alexgibson: I'm not entirely confortable with using a third party service to deliver placeholder images (even though they are likely all safe-for-work 😄). It would be nice to incorporate some more Firefox/Mozilla specific imagery into the demo, if we have something?

src/patterns/organisms/hero/hero-branded.hbs
```diff
@@ -0,0 +1,33 @@
...
+notes: |
+  - Adding a theme class of `mzp-t-firefox` will change the title font to Open Sans and add a default curve shape to the bottom.
+  - Product icons can be displayed using an additional theme class:
+    - `mzp-t-product-firefox`.
+    - `mzp-t-product-beta`.
+    - `mzp-t-product-developer`.
+    - `mzp-t-product-nightly`.
+    - `mzp-t-product-focus`.
+  - The bottom curve shape can be swapped with a different shape in your page’s CSS (it’s an SVG mask). You can also change the shape’s fill color in CSS with the selector `.mzp-c-hero.mzp-t-firefox:after`
+  - The page section immediately after/below the hero should have a background color matching the color of the curve shape so the sections will merge seamlessly.
```
alexgibson: Is this something that may be worth doing in the demo page, to show an example of intended usage? The abrupt end to the curves looks a little odd.

src/assets/sass/protocol/components/_hero.scss
```diff
@@ -0,0 +1,259 @@
...
+
+    // Dark theme
+    &.mzp-t-dark,
+    .mzp-t-dark & {
+        background-color: $color-gray-80;
+        color: $color-white;
+
+        .mzp-c-hero-desc,
+        .mzp-c-hero-tagline {
+            color: transparentize($color-white, .1);
```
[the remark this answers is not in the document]
craigcook: Thanks for reminding me... I wanted to run this by the design folks. The mockups have this text as pure white on a dark background, but we're using a dark gray on light backgrounds. So I tried a light gray for the dark theme, but when experimenting with different background colors (blue, purple, etc, and the Pocket page has a teal background) a neutral gray just doesn't look very good. So if we wanted this to be not-quite-white I think we should use a slight transparency rather than gray, or else just go with pure white for all text when the background is dark. We can discuss it or maybe the design team already discussed it and chose white.

src/assets/sass/protocol/components/_hero.scss
craigcook: Yeah, been meaning to bikeshed this in a meeting but it hasn't come up... I just don't see the benefit of having three different ways to declare the same value. I think we borrowed the separate `$padding-` and `$margin-` tokens idea from another system where they actually equaled different values, e.g. `$padding-sm` was 4px but `$margin-sm` was 8px or something like that. But we're not doing it that way so it just seems redundant in our case.

I can be persuaded if there's a benefit I'm missing, but I think I'd prefer just `$spacing-` for small spaces and `$layout-` for large spaces.

src/pages/demos/hero.hbs
craigcook: I was using placeimg as an easy way to test lots of different sizes and aspect ratios and intended to replace them with suitable local images, but was rushing to get this PR ready for review before I go missing for 36 hours. But I can swap in different images this morning.

src/assets/sass/protocol/components/_hero.scss
alexgibson: Sounds good to me, but we should make sure we file an issue and remove the tokens if everyone agrees 👍

src/patterns/organisms/hero/hero-branded.hbs
craigcook: Good call. I'll add some filler blocks to demonstrate the effect. (I did have some at one point; the demo CSS even still has the class because I was sloppy 😏 ).

src/assets/sass/protocol/components/_hero.scss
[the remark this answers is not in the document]
craigcook: Solid white it shall be.

craigcook:
Updated!

PR #191, opened 2018-07-30
Ensure documented publishing steps include updating dependencies

alexgibson:
It occured to me that there was a hole in our publishing steps for NPM, which did not cover someone's local NPM dependencies being out of date. This could result in unintentional publishing of a version of protocol-assets to protocol-tokens which does not match up to the version number in `package.json` for the `master` branch. This could happen quite easily when switching between feature branches, for example.

This PR updates the NPM `pretest` script to always do an `npm install` before `npm test` is run, so the `./dist` folder should always include the expected assets.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# HEAD
+
+## Bug Fixes
+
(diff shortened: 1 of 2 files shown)


alexgibson:
Rebased

PR #193, opened 2018-07-30
Move Jenkins notifications from IRC to Slack

pmac:


pmac:
This is regarding https://github.com/mozmeao/infra/issues/902

Issue #137, opened 2018-06-04
Update hover state for Card component


vincejoyiv:
Hey @alexgibson just noticing a couple small things on the card hoverstate.

- the box-shadow color should be "Gray 20" the hex value of which is `#EDEDF0`
- I'd like to do a border-bottom of 2px instead of 1.

That ends up looking like [this](https://www.dropbox.com/s/vbgtkh06xb2epvo/Screenshot%202018-06-04%2009.16.15.png?dl=0).

alexgibson:
This was done in #109

PR #197, opened 2018-08-02
Add inline CTA to card

### LINKED ISSUE #195: Add inline CTA to card component
To support the new about page design:

screen shot 2018-08-01 at 10 16 55

stephaniehobson:
Fixes #195

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
-# HEAD
-
 ## Features
 
+# HEAD
+
+* **docs:** Add inline CTA to card (#197)
 * **docs:** Add bidi control to component previews (#150)
 
 # 2.1.1
--- a/src/assets/sass/protocol/components/_card.scss
+++ b/src/assets/sass/protocol/components/_card.scss
@@ -70,10 +70,19 @@
     .mzp-c-card-title {
         @include text-display-sm;
         display: inline;
-        border-bottom: 1px solid transparent;
+    }
+
+    .mzp-c-card-title,
+    .mzp-c-card-cta-text {
+        border-bottom: 2px solid transparent;
     }
 
     .mzp-c-card-desc {
+        margin: $margin-xs 0 0;
+    }
+
+    .mzp-c-card-cta {
+        font-weight: bold;
         margin: $margin-xs 0 0;
     }
 
@@ -93,8 +102,9 @@
             @include transition(box-shadow .1s ease-in-out);
             box-shadow: 0 0 0 4px $color-gray-20;
 
-            .mzp-c-card-title {
-                @include transition(border-bottom .1s ease-in-out);
+            .mzp-c-card-title,
+            .mzp-c-card-cta-text {
+                @include transition(border-bottom-color 100ms ease-in-out);
                 border-bottom: 2px solid $color-black;
             }
 
@@ -144,11 +154,12 @@
 }
 
 /* -------------------------------------------------------------------------- */
-// Card with a white background & border for placement on colored backgrounds.
+// Card with extra small title and text
 
 .mzp-c-card.mzp-c-card-extra-small {
 
-    .mzp-c-card-tag {
+    .mzp-c-card-tag,
+    .mzp-c-card-cta {
         @include text-body-xs;
     }
 
--- a/src/patterns/molecules/card/small-card.hbs
+++ b/src/patterns/molecules/card/small-card.hbs
@@ -22,6 +22,7 @@ links:
       <div class="mzp-c-card-tag">Card tag</div>
       <h2 class="mzp-c-card-title">This example headline has 40 characters</h2>
       <p class="mzp-c-card-desc">A description with a maximum of 140 characters. That means we usually only have room for one or two sentences. Here is what that looks like.</p>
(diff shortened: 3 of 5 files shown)


stephaniehobson:
Demo: https://demo1--mozilla-protocol.netlify.com/patterns/molecules/card.html#medium-card-with-inline-text-cta

src/assets/sass/protocol/components/_card.scss
```diff
@@ -77,13 +77,19 @@
...
+    }
+
     .mzp-c-card-block-link {
         border-radius: 2px;
         display: block;
 
         &:link,
         &:visited {
             color: $color-black;
+            border-bottom: 1px solid transparent;
```
alexgibson: Should this be on `.mzp-c-card-title` and `.mzp-c-card-cta-text` instead of the block link? It should also be `2px` to match the hover state?

```css
.mzp-c-card-title,
.mzp-c-card-cta-text {
    border-bottom: 2px solid transparent;
}
```

src/patterns/molecules/card/medium-card-cta.hbs
```diff
@@ -0,0 +1,12 @@
...
+description: A medium card with image, headline, description, link, and inline text CTA
+order: 3
+notes: |
+    - Inline text CTAs should only be used in cards with descriptions
+links:
+    Card Layout Class: /patterns/templates/card-layout.html
+    Card Layout Demo: /demos/card-layout.html
+---
+
+{{#embed "patterns.molecules.card.small-card" class="mzp-c-card-medium mzp-has-aspect-3-2 mzp-has-video" cta="true" }}{{/embed}}
```
alexgibson: There's nothing intrinsically unique about this card type itself, so i'm not sure if it warrants its own handle-bars template & example here? Seeing as this inline CTA is an optional feature that can be used on any card type, and not something unique to the medium sized card, it may just be worth adding a note about it in the documentation. Or as an example to one of the card types on the existing demo page perhaps?

src/assets/sass/protocol/components/_card.scss
```diff
@@ -144,11 +151,12 @@
 }
 
 /* -------------------------------------------------------------------------- */
-// Card with a white background & border for placement on colored backgrounds.
+// Card with extra small title and text
```
alexgibson: Good catch, this was carry over from the first iteration of cards 👍

src/assets/sass/protocol/components/_card.scss
stephaniehobson: It's 1px to match what's on the cta link component. There's a transition that makes it look like it's growing.

And that transition means that it can't be put on a block level element :/ Personally, I'm not sure we should have custom underline effects with transitions. But since we do I'm trying to replicate how it was done.

Maybe it needs a mixin instead of being copied?

alexgibson: I think that 1px transition may be a mistake tbh, I don’t think it’s intentional?

src/patterns/molecules/card/medium-card-cta.hbs
stephaniehobson: I've moved it into the medium card.

src/assets/sass/protocol/components/_card.scss
stephaniehobson: Well, they should match. 

 @craigcook I know the discussion about the link underline effect was long and complicated. In https://github.com/mozilla/protocol/blob/master/src/assets/sass/protocol/base/elements/_links.scss#L28-L59 the transition is only on the colour and not the border-width. Should we remove the 1px border width? Increase it to 2px to match the hover state?

alexgibson: Fwiw, I think I wrote this code originally. It should be 2px, I just forgot to update the before-hover value when updating based on feedback from our designers. I think it’s safe just update it across the board and save a long discussion 🙈

alexgibson: Here’s the request to change it to 2px for transparency: https://github.com/mozilla/protocol/issues/137/

stephaniehobson:
Ready for re-review.

CHANGELOG.md
```diff
@@ -1,9 +1,3 @@
-# HEAD
```
alexgibson: Looks like this got removed by mistake in a rebase. We should also add this change to the changelog

src/assets/sass/protocol/components/_card.scss
```diff
@@ -70,20 +70,26 @@
...
+    }
+
     .mzp-c-card-block-link {
         border-radius: 2px;
         display: block;
 
         &:link,
         &:visited {
             color: $color-black;
+            border-bottom: 2px solid transparent;
```
alexgibson: This doesn't seem to be required (unless I'm missing something, which is also likely)

src/patterns/molecules/card/small-card.hbs
```diff
@@ -1,7 +1,7 @@
 ---
 name: Small Card
 description: A small card with image, headline, description, and link.
-order: 3
+order: 4
```
alexgibson: Is this needed still?

CHANGELOG.md
stephaniehobson: Wow, it managed to remove bits of both the PRs involved in the rebase.

stephaniehobson:
Changes made.

PR #205, opened 2018-08-10
Bump to v2.1.2

alexgibson:


alexgibson:
This is just updating the version number prior to publishing, so merging.

PR #208, opened 2018-08-14
Billboard alignment broken in Safari (Fixes #207)

### LINKED ISSUE #207: Billboard alignment broken in Safari (wide viewports)
As reported by @pmac - billboard components are showing mis-aligned content at wider viewports in Safari. I can reproduce this both on iPad and on Safari desktop.

file

Trying to degub a bit, it seems like Safari doesn't agree with the use of flexbox to vertically center the text, whilst still obeying the container left/right margin values.

alexgibson:
Fixes #207

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# HEAD
+
+## Bug Fixes
+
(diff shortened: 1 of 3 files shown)


alexgibson:
Thanks @stephaniehobson 🍻

PR #211, opened 2018-08-21
[fix #198] Update base button style, add dark theme

### LINKED ISSUE #198: New CTA button style
First time I've seen the [Pocket page](https://www.mozilla.org/en-US/firefox/pocket/) today, but I notice there's a new button style that's not yet in Protocol. 

We should add this - It seems like a new CTA button style perhaps (with unique hover style?), or could this be a new default button style for dark backgrounds?

craigcook:
Updates our base button style to match what we've been using in bedrock. This could conflict with sites using Protocol in production so this also bumps the version to 2.2.1.

Demo: https://demo1--mozilla-protocol.netlify.com/patterns/atoms/buttons.html

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "protocol",
-  "version": "2.2.0",
+  "version": "2.2.1",
   "private": true,
   "author": "Mozilla",
   "description": "A design system for Mozilla's websites.",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
-# HEAD
+# 2.2.1
+
+## Features
+
+* **css:** Update base button style, add dark theme (#198)
 
 ## Bug Fixes
 
--- a/src/assets/sass/docs/site.scss
+++ b/src/assets/sass/docs/site.scss
@@ -154,6 +154,11 @@
         border-top: 1px solid rgba(0, 0, 0, .2);
         margin: 30px -20px -10px;
     }
+
+    .mzp-t-dark {
+        background-color: $color-gray-60;
+        padding: $spacing-lg;
+    }
 }
 
 .protosite-pattern-code {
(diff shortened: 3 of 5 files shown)


alexgibson:
A few minor UI questions:

1. When hovered the button shadow pushes right up next to the neighbouring element unless there's a little extra padding. Maybe we're ok with this, but just flagging for discussion.

screen shot 2018-08-28 at 10 04 25

2. The dark-themed button hover state is pretty hard to see as the shadow is also dark. Could be an a11y issue?

screen shot 2018-08-28 at 10 04 35

3. Is the shadow supposed to be on download buttons as well?

screen shot 2018-08-28 at 10 04 45

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "protocol",
-  "version": "2.2.0",
+  "version": "2.2.1",
```
alexgibson: If you would like to publish this straight away, then you'd need to follow [these steps](https://github.com/mozilla/protocol/tree/master/docs#publishing) in addition to updating the version number here.

If not, just adding to the changelog under `HEAD` is fine, and we can update the version number when ready to next publish.

alexgibson: > This could conflict with sites using Protocol in production so this also bumps the version to 2.2.1.

If this is something that could conflict in production, we should also make it a major version bump according to https://semver.org/. I'm not sure what is a breaking change here though (unless I'm missing something).

craigcook: There are a few pages in bedrock with these same rules applied to buttons that override the older Protocol style, so I don't *think* this will cause major breakage but would at least make those overriding styles redundant. I can  do some testing to assess the potential for damage.

alexgibson: Ah, makes sense - then yes I'd say at least make this `2.3.0`. A patch version change would be picked up and installed via NPM automatically, which is probably not what we'd want here.

craigcook: Ah didn't realize minor versions were published automatically. I'll update.

craigcook:
Buttons have changed again so rather than land this I'm going to find time to work on updating it to the latest style.

craigcook:
Closing in favor of #224

PR #215, opened 2018-09-04
Add Navigation organism and Menu molecule (Fixes #182)

### LINKED ISSUE #182: Create Menu molecule & Navigation organism
Creatting this issue to track work for the new navigation component.

Sketch file plus wireframes: https://share.goabstract.com/967ec565-e446-4b30-aa03-65b3dd169d36

Content & structure: https://docs.google.com/document/d/1NNqfr5vtKj1tDz7HSBPTlWI2HFcXks7twzfZYLoK-hg/edit?usp=sharing

alexgibson:
Fixes #182

This has been pretty well tested during development, but could do with some sanity checking during code review. Main things to ensure are:

- [x] Navigation works on wide viewports.
- [x] Navigation works on small viewports.
- [x] Navigation is keyboard accessible.
- [x] Navigation still works with JS disabled.
- [x] RTL styles are displayed correctly.

Demo URLs:
- https://demo2--mozilla-protocol.netlify.com/demos/navigation.html
- https://demo2--mozilla-protocol.netlify.com/patterns/organisms/navigation.html
- https://demo2--mozilla-protocol.netlify.com/patterns/molecules/menu.html


alexgibson:
Ok this is not quite ready yet, need to refactor something

alexgibson:
Ok, ready.

src/assets/js/protocol/protocol-menu.js
```diff
@@ -0,0 +1,233 @@
...
+        e.preventDefault();
+        Menu.toggle(e.target.parentNode);
+    };
+
+    Menu.isWideViewport = function() {
+        return _mqWideNav.matches;
+    };
+
+    Menu.handleState = function() {
+        _mqWideNav = matchMedia('(min-width: 768px)');
```
stephaniehobson: Breakpoints should be global variables.

src/assets/js/protocol/protocol-menu.js
```diff
@@ -0,0 +1,233 @@
...
+
+        if (Menu.isWideViewport()) {
+            Menu.bindEventsWide();
+        } else {
+            Menu.bindEventsSmall();
+        }
+    };
+
+    Menu.bindEventsWide = function() {
+        var items = document.querySelectorAll('.mzp-c-menu-category.mzp-has-drop-down');
```
stephaniehobson: I know you're sick of looking at this....

But what do you think of using .mzp-js- classes with no styles on them to handle all behaviour? Is it worth doing the extra abstraction to we can use it with dramatically different styles? 

I'm thinking of the footer, but that will probably not have significant style differences. Not sure where else we might want to use this.

src/assets/js/protocol/protocol-menu.js
alexgibson: I don't think it needs to be global, but I can certainly move it to a variable.

src/assets/js/protocol/protocol-menu.js
alexgibson: Good call, something like `.mzp-js-expandable` may be appropriate perhaps. The menu does work without JS enabled, but I think just for purposes of toggling state (as opposed to styling), this would be ok.

alexgibson: It does mean we now have two classes (one for JS behavior and one for styling) which is more verbose, but if we want to keep things totally separate then fine by me.

stephaniehobson:
People navigating by keyboard often expect that ESC will close an open drop down, I noticed that is not the case for this one. Would it be difficult to add that functionality?

stephaniehobson:
I can't figure out what the icon next to Get Involved is. Popcorn? Two hands raised above a baseball cap?

stephaniehobson:
With javascript disabled on mobile the menu is expanded by default. That's a lot of scrolling to get to content. This is a pretty small edge case though and probably a fair amount of work to come up with a better solution?

Possible solutions:
- Move the menu to the bottom of the page
- Collapse the panels by default and only open the one corresponding to the section the user is currently inside of
- Only display the top level navigation items and rely on in-page links to access sub-section content.

stephaniehobson:
Content overflowing the container on a 320px wide screen:

screen shot 2018-09-10 at 10 24 20 am

stephaniehobson:
The + → x animation is sweet.

stephaniehobson:
At 768px in German the About menu is likely to wrap. This makes it difficult to get to the About option (but not impossible).

screen shot 2018-09-10 at 10 31 32 am

Possible solutions:
- It's fine and don't need any changes
- Hide the download button or wrap it under the Mozilla logo

stephaniehobson:
I think the mobile menu icon should get a pointer cursor.

Also, the user research on hamburger menus shows that adding a "menu" label or even just an outline around the icon will increase usage. https://conversionxl.com/blog/testing-hamburger-icon-revenue/

alexgibson:
> People navigating by keyboard often expect that ESC will close an open drop down, I noticed that is not the case for this one. Would it be difficult to add that functionality?

Good call, will add 👍 

> I've never really figured out why some interactive components get the pointing finger icon on hover and others don't so it feels weird to me that the pointer doesn't show up when hovering over the top level menu items. Is it worth changing or is expected browser behaviour and we shouldn't mess with it?

Will give this some thought.

> With javascript disabled on mobile the menu is expanded by default. That's a lot of scrolling to get to content. This is a pretty small edge case though and probably a fair amount of work to come up with a better solution?

I don't think we need to fix or change this now. As you say it's a pretty small edge case. My main concern was making sure links we're still accessible, which they are.

> Content overflowing the container on a 320px wide screen:

Will take a look at this, thanks.

stephaniehobson:
The icons seem a little low on mobile compared to desktop:

screen shot 2018-09-10 at 11 00 05 am

stephaniehobson:
Apparently I've volunteered myself to review :P

@craigcook You can still take a look from a Protocol perspective though if you want.

stephaniehobson:
On mobile when a second panel is opened the first does not close. When the second one is closed both the first and the second close. I think the panels controls should either be independent of each other for all actions or linked for all actions. 

Related (still on mobile): After opening 3 menus and then closing the 3rd the navigation menu is out of sight. 

Perhaps closing any panel should scroll the page back up to the top?

src/assets/sass/protocol/components/_menu.scss
```diff
@@ -0,0 +1,327 @@
...
+                z-index: 1001;
+            }
+
+            .mzp-c-menu-panel {
+                display: block;
+                z-index: 1000;
+            }
+        }
+
+        &:focus-within {
```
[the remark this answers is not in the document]
stephaniehobson: 🤩

alexgibson:
> On mobile when a second panel is opened the first does not close. When the second one is closed both the first and the second close. I think the panels controls should either be independent of each other for all actions or linked for all actions.

Ah great catch. The menus on mobile should be independent of each other. This is a bug that snuck in during my last refactor. Will fix.

src/assets/sass/protocol/components/_menu.scss
```diff
@@ -0,0 +1,327 @@
...
+        }
+    }
+
+    &.mzp-has-drop-down .mzp-c-menu-title {
+        cursor: default;
+        text-decoration: none;
+
+        &:before {
+            // base64 encoding of https://github.com/mozilla/protocol-assets/blob/master/icons/ui/expand-black.svg
+            background:  url('data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMTYiIGhlaWdodD0iMTYiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyI+PHBhdGggZD0iTTcgN1YxYTEgMSAwIDEgMSAyIDB2Nmg2YTEgMSAwIDEgMSAwIDJIOXY2YTEgMSAwIDEgMS0yIDBWOUgxYTEgMSAwIDEgMSAwLTJoNnoiIGZpbGw9IiMwMDAiIGZpbGwtcnVsZT0iZXZlbm9kZCIvPjwvc3ZnPgo=') top left no-repeat;
```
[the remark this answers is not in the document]
alexgibson: My main line of reasoning here was so the icons are visible immediately when the menu is opened, and don’t require an http request. I think making them immediately obvious in the UI is perhaps more advantageous than avoiding the debt, but i’m open to other opinions.

src/assets/sass/protocol/components/_menu.scss
```diff
@@ -0,0 +1,327 @@
...
+    background: $color-white;
+    display: none;
+    padding-top: $spacing-lg;
+    width: 100%;
+
+    // Close button is only vivible for large screens.
+    .mzp-c-menu-button-close {
+        @include image-replaced;
+        @include transition(transform 100ms ease-in-out);
+        background: transparent url('#{$image-path}/icons/ui/close-black.svg') center center no-repeat;
```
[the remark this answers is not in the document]
alexgibson: This seems like a less critica piece of UI, and not something that requires being loaded instantly when the panel opens. I’m happy to defer its visibility by a few milliseconds, unless you feel differently.

src/assets/sass/protocol/components/_menu.scss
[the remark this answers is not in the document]
craigcook: Yeah, that makes sense. And in the case of mobile I agree, the benefit of instantly available outweighs the cost of easily changeable (when it might never even change).

src/patterns/organisms/navigation/navigation.hbs
```diff
@@ -0,0 +1,37 @@
...
+    <div class="mzp-c-navigation-container">
+      <button class="mzp-c-navigation-menu-button" type="button" aria-controls="map-c-navigation-items">Menu</button>
+      <div class="mzp-c-navigation-logo"><a href="https://www.mozilla.org/">Mozilla</a></div>
+      <div class="mzp-c-navigation-items" id="map-c-navigation-items">
+        <div class="mzp-c-navigation-items-container">
+          {{#block "menu"}}
+            {{#embed "patterns.molecules.menu.menu"}}{{/embed}}
+          {{/block}}
+          <div class="mzp-c-navigation-download">
+            {{#embed "patterns.molecules.download-button.download-button" class="mzp-t-download-secondary"}}
```
stephaniehobson: The download button doesn't work. Maybe the molecule is included just for demo purposes and this doesn't reflect bedrock functionality.

stephaniehobson:
Tested in:
- Firefox
- iPad portrait
- iPad landscape

stephaniehobson:
Both the menu `li` and  `a` are having the `aria-expanded` attribute applied to them and only the `li` is having it toggled when the menu is opened/closed. Leading to the rather confusing announcement: "About. Link. Collapsed. Has popup." when the menu is open :/

Also, the `li` is missing the `aria-controls` that needs to be paid with `aria-expanded`. Probably removing the aria elements from the `li` will fix the problem.

stephaniehobson:
Not a blocker: The nav items are not vertically centered in IE 11.
screen shot 2018-09-10 at 3 42 46 pm

src/assets/js/protocol/protocol-menu.js
```diff
@@ -0,0 +1,311 @@
...
+     * @param {Boolean} animate - show animation when menu panel opens.
+     */
+    Menu.open = function(el, animate) {
+        if (animate) {
+            el.classList.add('animate');
+        }
+
+        el.classList.add('selected');
+        el.setAttribute('aria-selected', true);
+        el.setAttribute('aria-expanded', true);
```
stephaniehobson: I don't think we need both aria-selected and aria-expanded together.

tests/unit/protocol-menu.js
```diff
@@ -0,0 +1,182 @@
...
+    'use strict';
+
+    beforeEach(function () {
+        var menu = '<nav class="mzp-c-menu">' +
+                     '<ul class="mzp-c-menu-category-list">' +
+                       '<li class="mzp-c-menu-category mzp-has-drop-down mzp-js-expandable">' +
+                          '<a class="mzp-c-menu-title" href="#" aria-haspopup="true" aria-controls="mzp-c-menu-panel-example">Firefox</a>' +
+                          '<div class="mzp-c-menu-panel mzp-has-card" id="mzp-c-menu-panel-example">' +
+                            '<div class="mzp-c-menu-panel-container">' +
+                              '<button class="mzp-c-menu-button-close" type="button" aria-controls="mzp-c-menu-panel-example">Close</button>' +
```
stephaniehobson: The `aria-controls` needs to match the ID of the menu it controls.

src/assets/js/protocol/protocol-menu.js
```diff
@@ -0,0 +1,311 @@
...
+
+        Menu.close();
+    };
+
+    /**
+     * Toggles the open/closed state of a menu panel.
+     * @param {Object} el - DOM element (`.mzp-c-menu-category.mzp-js-expandable`)
+     */
+    Menu.toggle = function(el) {
+        var state = el.classList.contains('selected') ? true : false;
```
stephaniehobson: `.is-selected`?  I think this is the same element that Craig commented on in his review.

src/assets/js/protocol/protocol-menu.js
```diff
@@ -0,0 +1,311 @@
...
+        if (!state) {
+            Menu.open(el);
+        } else {
+            Menu.close();
+        }
+    };
+
+    /**
+     * Menu `mouseenter` event handler.
+     * Opens the menu only when hover intent is shown.
```
stephaniehobson: 💯

src/assets/js/protocol/protocol-menu.js
```diff
@@ -0,0 +1,311 @@
...
+     * Menu `focusout` event handler.
+     * Closes the menu when focus moves to an alement outside of the currently open panel.
+     */
+    Menu.onFocusOut = function() {
+        var self = this;
+
+        /**
+         * After an element loses focus, `document.activeElement` will always be `body` before
+         * moving to the next element. A `setTimeout` of `0` circumvents this issue as it
+         * re-queues the JavaScript to run at the end of the current excecution.
```
stephaniehobson: That must have been a PITA to figure out.

src/assets/js/protocol/protocol-menu.js
```diff
@@ -0,0 +1,311 @@
...
+        return typeof window.matchMedia !== 'undefined' &&
+               window.matchMedia('all').addListener &&
+               'classList' in document.createElement('div');
+    };
+
+    /**
+     * Initialise menu.
+     * @param {Object} options - configurable options.
+     */
+    Menu.init = function(options) {
```
stephaniehobson: I don't see where this is called, except in the tests. Is this function just here for the tests?

src/assets/js/protocol/protocol-navigation.js
```diff
@@ -0,0 +1,82 @@
...
+        e.preventDefault();
+
+        // Update button state
+        e.target.classList.toggle('active');
+
+        // Update menu state
+        menu.classList.toggle('open');
+
+        // Update aria-expended state on menu.
+        var expanded = menu.classList.contains('open') ? true : false;
```
stephaniehobson: `.is-open`?

stephaniehobson:
I feel like a bit of an idiot asking but... how do the menu and navigation get initialized? They're clearly working but I only see them called in the tests.

The line between the menus and the navigation JS seems really blury right now, they both seems to be handling click events and aria events. Maybe we can sync on this in the morning for you to give me an overview.

src/patterns/organisms/navigation/navigation.hbs
alexgibson: Yeah the download button in bedrock is way more complex. Right now, Protocol is only deals with how it should appear visually. Some day it would be good to close the gap here, but that's another project itself :)

alexgibson:
@stephaniehobson thanks so much for the thorough code review 💯 

I added a second commit with some review fixes. Summary here:

- Update `animate` and `selected` classes to `is-animated` and `is-selected`.
- Close menu with `esc` key press on wide viewports.
- Ensure `Menu.Toggle` only closes the currently selected menu panel.
- Fix menu panel overflow on small viewports.
- Shrink Mozilla logo and hide download button on menium viewports to maximise nav space for l10n.
- Fix aria states and improve close button descriptive text.

src/pages/demos/navigation.hbs
```diff
@@ -0,0 +1,430 @@
...
+</div>
+
+<script src="{{@root.baseurl}}/assets/protocol/protocol/js/protocol-menu.js"></script>
+<script src="{{@root.baseurl}}/assets/protocol/protocol/js/protocol-navigation.js"></script>
+
+<script>
+(function() {
+    'use strict';
+
+    Mozilla.Navigation.init({
```
alexgibson: @stephaniehobson this is where init gets called in the demo page. It’s also detailed in the docs for both components.

stephaniehobson: Oooooooh. That explains why I coulldn't find it in the .js files >_<

alexgibson:
@stephaniehobson pushed another commit so the links naturally wrap underneath the logo + download button, should the menu length exceed the space inbetween. Thanks for your suggestion 👍

stephaniehobson:
New wrapping works like a charm ✨

src/assets/sass/protocol/components/_navigation.scss
```diff
@@ -58,79 +68,74 @@
...
-        .mzp-c-menu {
-            @include bidi(((float, left, right),));
-        }
-
-        .mzp-c-navigation-download {
-            display: none;
-        }
+        display: block;
+        margin: $spacing-lg 0;
+        max-width: 180px;
```
stephaniehobson: The wrapping works really well without a max-width on the button.

Not a blocker though.

src/assets/js/protocol/protocol-navigation.js
```diff
@@ -0,0 +1,82 @@
...
+     */
+    Navigation.setAria = function() {
+        menu.setAttribute('aria-expanded', false);
+    };
+
+    /**
+     * Bind navigation event handlers.
+     */
+    Navigation.bindEvents = function() {
+        menu = document.querySelector('.mzp-c-navigation-items');
```
stephaniehobson: Do you think we could we name this variable something other than menu? Maybe nav-items?

alexgibson:
https://media.giphy.com/media/5mBE2MiMVFITS/giphy.gif

PR #223, opened 2018-09-11
Font Optimization

### LINKED ISSUE #219: Font Optimization
`-moz-osx-font-smoothing: grayscale`
This is a CSS property that will only be applied to websites rendered through Firefox on a MacOS device. I tested this on a 2014 non-retina Macbook Air and a 2018 retina MacBook Pro and it helped with legibility so I think we should include this property.

`-webkit-font-smoothing: antialiased`
This is a property that Google applies to Open Sans on their websites (google.com/fonts/specimen/Open+Sans). So I lean towards including it because it seems as if the original intent for this font is for font-smoothing to be applied. But I'm all for holding off until we can do some testing.

michaelmccombie:
A couple of quick fixes to make our fonts easier to read. Resolves #219

### DIFF
--- a/src/assets/sass/protocol/base/elements/_document.scss
+++ b/src/assets/sass/protocol/base/elements/_document.scss
@@ -15,6 +15,9 @@ body {
     background: #fff;
     color: $color-black;
     font-family: $base-font-stack;
+    -moz-osx-font-smoothing: grayscale;
+    -webkit-font-smoothing: antialiased;
+    text-rendering: optimizeLegibility;
 }
 
 // A few places use a class to hide elements. We should try to avoid that.
--- a/src/assets/sass/protocol/base/elements/_document.scss
+++ b/src/assets/sass/protocol/base/elements/_document.scss
@@ -17,7 +17,6 @@ body {
     font-family: $base-font-stack;
     -moz-osx-font-smoothing: grayscale;
     -webkit-font-smoothing: antialiased;
-    text-rendering: optimizeLegibility;
 }
 
 // A few places use a class to hide elements. We should try to avoid that.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Bug Fixes
 
+* **css:** Add font-smoothing properties for macOS devices
 * **npm:** Ensure `npm test` always updates dependencies before running `gulp build`.
 
 # 2.2.0


src/assets/sass/protocol/base/elements/_document.scss
```diff
@@ -15,6 +15,9 @@ body {
     background: #fff;
     color: $color-black;
     font-family: $base-font-stack;
+    -moz-osx-font-smoothing: grayscale;
+    -webkit-font-smoothing: antialiased;
+    text-rendering: optimizeLegibility;
```
alexgibson: Before applying non-standard properties like these, we should pause to consider how each operating system treats them. Something that looks better on Mac, may not necessarily look better on Windows for example.

In the case of `text-rendering`, there are also some pretty serious performance penalties that can occur on mobile devices, should there be a lot of text on a page https://css-tricks.com/almanac/properties/t/text-rendering/

alexgibson: Further reading. I’m not saying this change is wrong, but it may be more heavy handed than it first appears. 

http://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/

michaelmccombie: `-moz-osx-font-smoothing: grayscale` 
This is a CSS property that will only be applied to websites rendered through Firefox on a MacOS device. I tested this on a 2014 non-retina Macbook Air and a 2018 retina MacBook Pro and it helped with legibility so I think we should include this property.

`-webkit-font-smoothing: antialiased`
This is a property that Google applies to Open Sans on their websites (google.com/fonts/specimen/Open+Sans). So I lean towards including it because it seems as if the original intent for this font is for font-smoothing to be applied. But I'm all for holding off until we can do some testing. 

`text-rendering: optimizeLegibility` 
I haven't tested this myself, but the article you linked to identifies enough performance issues that it's probably not worth including.

alexgibson: If we're confident that things will still remain readable on lower resolution screens, then the Mac-specific properties should be fine. We'll just need to be mindful when designing new components, as the gap between text rendering on Mac vs Windows will be larger.

+1 for removing `text-rendering`

Please make sure to add this PR to the changelog so we have a record if it 👍

PR #224, opened 2018-09-12
Update button styling

michaelmccombie:
A handful of design related changes here that myself and @vincejoyiv worked on recently.

- Adds 'disabled' state
- Adds a 'small' size for things like the header and forms
- Overhauls hover states
- Resizes the buttons
- Removed light shadows and gradients


alexgibson:
Note: This will probably need rebasing should #215 merge first, as it covers some of the same things.

src/assets/sass/protocol/components/_button.scss
```diff
@@ -4,36 +4,151 @@
...
     display: inline-block;
     font-weight: bold;
-    padding: $padding-md $padding-lg;
+    padding: calc(#{$padding-md} - 2px) calc(#{$padding-xl} - 2px); //2px extra padding removed to compensate for 2px border.
     text-decoration: none;
 
-    &:hover,
-    &:focus {
-        background-color: $color-white;
+    &:hover {
```
alexgibson: Need to retain the `focus` state here for keyboard users.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## Features
```
alexgibson: Please add this under `#HEAD`

src/assets/sass/protocol/components/_button.scss
```diff
@@ -4,36 +4,151 @@
...
+
 .mzp-t-dark .mzp-c-button {
     background-color: $color-white;
     border-color: $color-white;
     color: $color-black;
 
-    &:hover,
-    &:focus {
-        background-color: $color-gray-80;
+    &:hover {
```
alexgibson: Missing `focus`

src/assets/sass/protocol/components/_button.scss
```diff
@@ -4,36 +4,151 @@
...
+        color: $color-gray-30;
+    }
+}
+
+.mzp-c-button.mzp-t-download-secondary {
+    background-color: transparent;
+    border-color: $color-blue-60;
+    color: $color-blue-60;
+
+    &:hover {
```
alexgibson: Missing `focus`

src/assets/sass/protocol/components/_button.scss
```diff
@@ -4,36 +4,151 @@
...
+/* -------------------------------------------------------------------------- */
+// Download Buttons
+
+.mzp-c-button.mzp-t-download {
+    background-color: $color-blue-60;
+    border: none;
+    color: $color-white;
+    padding: $padding-md $padding-xl;
+
+    &:hover {
```
alexgibson: Missing `focus`

src/assets/sass/protocol/components/_button.scss
```diff
@@ -4,36 +4,151 @@
...
+        background-color: $color-blue-70;
+        color: $color-white;
+        box-shadow: none;
+        transform: none;
```
alexgibson: `transform` mixin

src/assets/sass/protocol/components/_button.scss
```diff
@@ -4,36 +4,151 @@
...
+    background-color: transparent;
+    border-color: $color-blue-60;
+    color: $color-blue-60;
+
+    &:hover {
+        background-color: $color-blue-70;
+        border-color: $color-blue-70;
+        color: $color-white;
+        box-shadow: none;
+        transform: none;
```
alexgibson: `transform` mixin

alexgibson:
Both page Hero and Call Out examples and demo pages need updated download buttons:

screenshot_2018-10-16 page hero - protocol design system


screenshot_2018-10-16 call out - protocol design system

src/assets/sass/protocol/components/_button.scss
michaelmccombie: @alexgibson I disagree. We really shouldn't be applying any custom focus states. By removing this selector we are preserving the default focus state. It's a best practices from a usability standpoint to keep the default state because users are already familiar with it and know to look for it when using a site. It's the same reason we don't customize the cursor on our site.

tabbing

src/assets/sass/protocol/components/_button.scss
```diff
@@ -4,36 +4,151 @@
...
-    &:focus {
-        background-color: $color-white;
+    &:hover {
+        background-color: transparent;
         color: $color-black;
+        box-shadow: 4px 4px 0 0 $color-black;
+        transform: translate(-2px, -2px);
+    }
+
+    &:disabled {
```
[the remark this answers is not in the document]
michaelmccombie: @alexgibson I included the disabled state because it was missing from our CSS file and having a button that is disabled but looks and functions like a non-disabled button is a really bad experience. 

But I think we should avoid disabled buttons whenever possible. A better way to handle things is to enable the button and give the user a notification when they can't continue because of missing information (this is how the sign up form works).

So maybe we shouldn't include an example. My hope is that if someone doesn't know its a possibility then they won't ask for it.

src/assets/sass/protocol/components/_button.scss
stephaniehobson: I used to think that too but I was told at the Hack on MDN Accessibility weekend (last month) that really visible focus styles are important. There was also [a great presentation about it](https://noti.st/ericwbailey/TcMJFP/if-it-s-interactive-it-needs-a-focus-style) at a11yToConf. 

I would not mess with the *outline* but dramatic `:focus` states improves the keyboard user experience. (A specific use case for these would be if it was appearing on a blue background.)

alexgibson: @michaelmccombie i’m not trying to suggest doing anything with `outline` here. The key thing to remember is that browsers all implement focus rings differently. Whilst browsers like Chrome show a bright blue focus ring like in your screenshot, other browser like Firefox use a smaller dotted black outline. Testing your branch out in Firefox, it’s almost impossible to tell when a button is focused currently.

As @stephaniehobson says, if it’s interactive it should really have a focus style. That can be anything we decide, but don’t mess with the default `outline`.

michaelmccombie: Sounds good to me, thanks Stephanie and Alex for looking in to this. I'll update things 👍

alexgibson:
The download button molecule shows the button is misaligned with the privacy link. The secondary download button example here may also now be redundant? (the styling is currently broken either way).

screen shot 2018-11-07 at 10 27 39

alexgibson:
The navigation demo is showing a full-fat download button:

screen shot 2018-11-07 at 10 28 16

src/assets/sass/protocol/components/_button.scss
```diff
@@ -4,36 +4,153 @@
...
 @import '../includes/lib';
 
+/* -------------------------------------------------------------------------- */
+// Primary Button
+
 .mzp-c-button {
     @include border-box;
     @include text-cta;
-    transition: background-color 100ms, border-color 100ms, color 100ms;
+    transition: background-color 100ms, box-shadow 100ms, color 100ms, transform 100ms;
```
alexgibson: Nit: use `transition` mixin. 

Alphabetising the CSS properties in here would also be nice for consistency (here and below).

src/patterns/atoms/buttons/secondary-dark.hbs
```diff
@@ -0,0 +1,33 @@
+---
+name: Secondary (Dark)
+order: 4
+description: |
+  This is a secondary button displayed on a dark background. Add the class `mzp-t-dark` to change the button color from dark to light.---
```
alexgibson: Remove `---` from end of text?

src/patterns/atoms/buttons/download-secondary.hbs
```diff
@@ -0,0 +1,36 @@
+---
+name: Download (Secondary)
+order: 6
+description: |
+  A product download button theme. This button is used when the download CTA is not the main action a person should take on a page.
```
alexgibson: `A product download secondary button theme` ?

src/patterns/atoms/buttons/download.hbs
```diff
@@ -1,32 +1,35 @@
 ---
-name: Download button
+name: Download
+order: 5
 description: |
   A product download button theme, with additional secondary variation.
```
alexgibson: As there's no longer a secondary example included here, remove `with additional secondary variation`?

src/assets/sass/protocol/components/_button.scss
alexgibson: Alphabetising the CSS properties in in this file would also be nice for consistency (both here and below), although not a blocker.

michaelmccombie:
@alexgibson Thanks for the feedback! This branch should be ready for review again.

PR #228, opened 2018-09-13
Fix card layout issues at medium viewport sizes (Fixes #227)

### LINKED ISSUE #227: Card layout bugs (medium viewports)
I noticed some layout bugs on the current home page. It only happens between 1010px and 1050px:

- Large card has an incorrect max-width.
- Small cards have an incorrect max width.
- Extra small (pocket cards) have an incorrect max-width.
- 6 card layout has a float clearance issue due to a long title and description.

Screenshot below:

screenshot_2018-09-13 internet for people not profit

alexgibson:
Fixes #227 

I verified this fix by looking at the card layout demo page, and also by changing these two values directly on the www.mozilla.org home page using dev tools.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 * **docs:** Add page that lists all tokens and their values (#177)
 
+## Bug Fixes
+
+* **css:** Fix card layout issues at medium viewport sizes (#227)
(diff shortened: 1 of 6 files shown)


alexgibson:
This is fixing a bug in production, so would be good to get reviewed & merged soon.

PR #231, opened 2018-09-14
Include Navigation in core bundle and not extras

alexgibson:
Well, I was bound to get something wrong 🤦‍♂️ 

This is a small change that pulls the navigation out of extras and into the core bundle (since it is a global component), and bumps the version number to 2.3.1.


src/assets/sass/protocol/base/_elements.scss
```diff
@@ -10,3 +10,4 @@
 @import 'elements/tables';
 @import 'elements/forms';
 @import 'elements/sections';
+@import 'elements/animations';
```
alexgibson: I know animations don't really tie directly to an element, but this felt like the most logical place to put this, without making a new directory under `/base/`. Happy to change this though

PR #236, opened 2018-09-20
Fix #234: Add blockquote

### LINKED ISSUE #234: Create a base style for blockquotes and (maybe) a component for pull quotes
The `blockquote` element currently receives no special styling, just resetting browser defaults so it looks like normal text. We need to design a general style for the `blockquote` element and possibly a fancier component for pull quotes, if we want to give them special treatment.

amychurchwell:
Styles `blockquote` element per example provided in issue #234

### DIFF
--- a/src/assets/sass/protocol/base/elements/_quotes.scss
+++ b/src/assets/sass/protocol/base/elements/_quotes.scss
@@ -1,0 +1,23 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import '../../includes/lib';
+
+
+blockquote {
+    @include bidi((
+        (border-left, 5px solid $color-gray-20, border-right, 0),
+        ));
+    @include text-display-md;
+    @include zilla-slab;
+    margin: $spacing-md auto;
+    padding: $spacing-sm $spacing-lg;
+    font-weight: bold;
+
+    & > :last-child {
+        margin-bottom: 0;
+    }
+}
+
+
--- a/src/patterns/atoms/typographic/quotes.hbs
+++ b/src/patterns/atoms/typographic/quotes.hbs
@@ -1,0 +1,14 @@
+---
+name: Quote Elements
+description: |
+  A pull quote is an excerpt that has been pulled from a body of text used to highlight a key topic. A blockquote is a quotation that is set off from the main text typically used for long quotations.
+links:
+  "MDN <blockquote>": https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blockquote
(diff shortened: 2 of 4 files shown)


src/patterns/atoms/typographic/quotes.hbs
```diff
@@ -0,0 +1,14 @@
+---
+name: Quote Elements
+description: |
+  A pull quote is an excerpt that has been pulled from a body of text used to highlight a key topic. A blockquote is a quotation that is set off from the main text typically used for long quotations.
+links:
+  "MDN <blockquote>": https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blockquote
+order: 5
+---
+
+<blockquote class="mzp-c-quote">
```
craigcook: This shouldn't have a class since right now we're just styling the generic element. If we later decide to make a special component for pull quotes or some other special quote, that would get a class to distinguish it from plain `blockquote` elements.

src/assets/sass/protocol/base/elements/_quotes.scss
```diff
@@ -0,0 +1,21 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import '../../includes/lib';
+
+
+blockquote {
+    border-left: 5px solid #ededf0;
```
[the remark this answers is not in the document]
craigcook: We should also use a color token here. This is `$color-gray-20`

src/patterns/atoms/typographic/quotes.hbs
```diff
@@ -0,0 +1,14 @@
...
+  <p>
+    "Chupadados is the 21st century iteration, more modern and more cute," Varon said. "It means 'date sucker,'"
```
craigcook: If you like, you can use `{{data "specimens.sentence"}}` to get a sample sentence here.
https://github.com/mozilla/protocol/blob/master/src/data/specimens.yaml

craigcook:
We should also update `CHANGELOG.md` and update the version number so we can publish this to NPM once it merges.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "protocol",
-  "version": "2.3.1",
+  "version": "2.3.2",
```
alexgibson: There are more things to bump than the version number here. I'd be happy to walk you through the release process, as it's not as simple as some npm packages.

alexgibson:
@amychurchwell please just rebase your branch against master and make sure your changelog entry appear underneath where it says `#HEAD`. Then I think we're good to merge.

@craigcook I think I'd like to discuss how & when we release to NPM, just so everyone understands the process. This is probably a good thing to walkthrough during our next Protocol meeting.

PR #239, opened 2018-09-24
Workaround for IE9 navigation bug in bedrock's classList polyfill (#182)

alexgibson:
This is a small bug fix to workaround a quirk in bedrock's `classList` polyfill for IE9. I filed a separate issue for bedrock to investigate updating our polyfill to the latest version on MDN:

https://github.com/mozilla/bedrock/issues/6221

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ## Bug Fixes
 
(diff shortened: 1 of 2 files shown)


alexgibson:
@craigcook I think Stephanie might be out, r on a bug fix for the nav?

PR #244, opened 2018-10-02
Add mixins and classes for the grid system

michaelmccombie:
This update adds new mixins and classes for the grid system. There is also some slight refactoring of older components to allow for the removal of old mixins. 

**New Mixins**
- `grid($number)` = sets the width of an element
- `overflow($number)` = positions a grid cell absolutely so it can flow past the grid container
- `overflow-sibling($number)` = adds a margin-left to make up for the abs positioning of a previous element with the overflow mixin

**New Classes**
- `mzp-l-grid` = equally divides all immediate child divs
- `mzp-l-grid-($number)-($number)` = an override for changing the width of an element within the grid
- `mzp-u-right` = floats content to the left
- `mzp-u-center` = removes floats and adds margin-auto
- `mzp-u-overflow-hidden` = Adds overflow-x: hidden


alexgibson:
I haven't delved into reviewing code here, but I did some preliminary browser testing and found some issues:

In FIrefox there are orphans at medium screen sizes (grid, three column and overflow)

firefox-medium-width

alexgibson:
In IE (tested down to version 9) there are also some orphans at desktop size:

There are also orphans in the main grid at desktop size in IE (I tested in 9):

screen shot 2018-10-04 at 12 19 55

alexgibson:
The article page demos also appear to be displaying at odd sizes:

screenshot_2018-10-04 article with sidebar menu - protocol design system
screenshot_2018-10-04 navigation - protocol design system

alexgibson:
Some other thoughts:

- This is currently listed under "Atoms" in the menu, but it's really a layout class. Would it be better housed under "Templates", along with the article and card layout classes?
- We should link to the grid demo page from the pattern.

michaelmccombie:
> I haven't delved into reviewing code here, but I did some preliminary browser testing and found some issues:
> 
> In FIrefox there are orphans at medium screen sizes (grid, three column and overflow)
> 

This is intentional and some of this layout will probably change when we finish up the second part of the grid update and come up with more tangible use cases. 

The reasons there are orphans is because we don't want give anything more visual weight than the items its displayed with. This is actually based on the way our card components work. If you had three card components displayed at this screen size they would have space to the right of the last card. We avoid this on the homepage by sticking to 2, 4, and 6 card pairings. 

I think this is something that could change though as we start to refactor things and think through more use cases.

image

michaelmccombie:
> In IE (tested down to version 9) there are also some orphans at desktop size:
> 
> screen shot 2018-10-04 at 12 19 55

Thanks for pointing this out. This is likely due to a rounding error with the `calc` property. Should be simple enough to fix by rounding down instead of including the fraction (which we've already done for a few other columns). I'll investigate further and push an update.

image

alexgibson:
> This is intentional and some of this layout will probably change when we finish up the second part of the grid update and come up with more tangible use cases.

Ok yeah, this makes sense. For card layout we avoid odd numbers of cards for this reason. Maybe we want to think about that here too in terms of guidelines.

As for the changes to the article width, we may want to think about how this might effect any pages in production (if any are using this).

michaelmccombie:
> As for the changes to the article width, we may want to think about how this might effect any pages in production (if any are using this).

It looks like two pages are currently using our article component:
- Email Privacy
- Profile-Per-Install

The Email Privacy pages have some small, non-consequential differences. The Profile-Per-Install pages have virtually no changes because the protocol classes are being overridden with a custom stylesheet.

article page

alexgibson:
Before we get into the weeds of code review for this PR, I'd be very much interested in thoughts from other front-enders on the overall approach here. Some high-level questions:

- Does creating a grid system like this address the immediate needs of Protocol? 
- Is a system like this doing too much, or is it well balanced enough? 
- Does the balance of utility classes vs the mixins feel right? 
- Should the focus be on layout, or just container sizes? (how does this fit into a future world where we use CSS grid).

/cc @craigcook @stephaniehobson @amychurchwell.

ejregithub:
Consensus is not to use a grid system  . . . Decided in the team DS meeting w. JGruen on 3/28

PR #248, opened 2018-10-05
Fixed menu component focusout bug (Fixes #247)

### LINKED ISSUE #247: Menu component focus bug
Spotted a late-breaking bug in the Menu component :(

STR:

1. Visit https://protocol.mozilla.org/demos/navigation.html
2. **Click** on "Firefox" in the top navigation menu.
3. **Hover** over to "About"
4. In the "About" menu, try and click on a link.

Expected results:

Page should navigate to the chosen destination.

Actualy results:

The menu closes and page does not navigate.

alexgibson:
Fixes #247

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 * **docs:** Document deployment process (#171)
 
+## Bug Fixes
+
+* **js:** Fixed menu component focusout bug (#247)
+
 # 2.4.2
 
 * **css:** Add blockquote component (#234)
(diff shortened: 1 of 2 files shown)


alexgibson:
Thanks @stephaniehobson - the bug occurs because when focus moves from within a menu panel to outside of it. Unless we check that the menu from which focus is moving _from_ is actually open, we can inadvertantly close the wrong panel (which is what this code change fixes). Hope that makes sense.

PR #260, opened 2018-10-31
Update Language Selector to include optional link to language selection page (Fixes #259)

### LINKED ISSUE #259: Update Language Selector to include optional link to a languages page
In https://github.com/mozilla/bedrock/issues/5827 we're building a new language selection page for bedrock, and the ask is to make the label "Language" in the Language Selector a link:

screen shot 2018-10-30 at 9 08 23 am

Not every site may have such a page, so we should probably make this optional:

- If no link is supplied, the "Language" `<label>` should be visible.
- If a link is supplied, the "Language" `<label>` should be visually hidden (still accessible by screen readers), and the link shown in its place.

alexgibson:
Fixes #259

### DIFF
--- a/src/assets/sass/protocol/components/_language-switcher.scss
+++ b/src/assets/sass/protocol/components/_language-switcher.scss
@@ -8,9 +8,16 @@
     color: inherit;
     font-weight: bold;
 
-    label {
+    label,
+    .mzp-c-cta-link {
         @include bidi(((margin, 0 $margin-lg $margin-sm 0, 0 0 $margin-sm $margin-lg),));
+        @include text-body-sm;
         display: inline-block;
+    }
+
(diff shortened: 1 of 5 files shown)


src/assets/sass/protocol/components/_language-switcher.scss
```diff
@@ -8,11 +8,18 @@
...
-    label {
+    label,
+    .mzp-c-cta-link {
         @include bidi(((margin, 0 $margin-lg $margin-sm 0, 0 0 $margin-sm $margin-lg),));
+        @include text-body-sm;
         display: inline-block;
     }
 
+    // hide the <label> visually should a language link be shown instead.
+    .mzp-c-cta-link + label {
```
stephaniehobson: 🔥

PR #305, opened 2019-02-11
Update ESLint to consume @mozilla-protocol/eslint-config (Fixes #85)

### LINKED ISSUE #85: Publish our shared ESLint and Stylelint configs
It would be nice if we could publish our ESLint and Stylelint configs, so both protocol and the desgin tokens could import the same files (as well as any other projects that may want a similar coding style)

alexgibson:
Updates Protocol to consume https://github.com/mozilla/eslint-config-protocol/

Fixes #85

Benefits of this change:

- Enables us to work from a shared ESLint config across multiple projects (I'm going to update the other Protocol repos to use this package, as well as bedrock).
- Creates separate configs for both [browser](https://github.com/mozilla/eslint-config-protocol/blob/master/index.js) and [node](https://github.com/mozilla/eslint-config-protocol/blob/master/index-node.js) related files for greater separation of concerns, plus more flexibility 🤓.

### DIFF
--- a/src/assets/js/protocol/protocol-menu.js
+++ b/src/assets/js/protocol/protocol-menu.js
@@ -7,7 +7,7 @@ if (typeof Mzp === 'undefined') { // eslint-disable-line block-scoped-var
     var Mzp = {};
 }
 
-(function() {
+(function(Mzp) {
     'use strict';
 
     var Menu = {};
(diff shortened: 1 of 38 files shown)


src/assets/js/protocol/protocol-menu.js
```diff
@@ -309,8 +309,8 @@ if (typeof Mzp === 'undefined') {
      * Basic feature detect for 1st class menu JS support.
      */
     Menu.isSupported = function() {
-        if (typeof Mzp.Supports !== 'undefined') {
-            return Mzp.Supports.matchMedia && Mzp.Supports.classList;
+        if (typeof window.Mzp.Supports !== 'undefined') {
```
stephaniehobson: Would it make sense to pass `window.Mzp` into the parent function as `Mzp`?

alexgibson:
@stephaniehobson thanks for the review. Updated and added to changelog 👍

Issue #304, opened 2019-02-11
Cursor pointer missing on navigation links without a hover menu


alexgibson:
If a navigation menu item does not have a `mzp-has-drop-down` or `mzp-js-expandable` class, then it acts like a regular link e.g.

```html
<li class="mzp-c-menu-category">
  <a class="mzp-c-menu-title" href="https://www.mozilla.org/firefox/">Firefox</a>
</li>
```

Whilst this works as intended, hovering over the link will not show the regular pointer cursor:

![menu-link](https://user-images.githubusercontent.com/400117/52554814-f0a7da80-2ddf-11e9-9f8f-1349fd652eab.gif)

This seems to be the offending line https://github.com/mozilla/protocol/blob/master/src/assets/sass/protocol/components/_menu.scss#L100

stephaniehobson:
Fixed in #308

Issue #300, opened 2019-02-08
Only one naviagation menu per page


stephaniehobson:
I'm trying to re-use the menu component and discovered a bug that only allows one per page.

This make a lot of sense... 

... but the fix is small so I will make it.

stephaniehobson:
Fixed in #325 #238 #306.

PR #306, opened 2019-02-12
Navigation bug fixes

stephaniehobson:
- Now supports multiple navigation menus on the same page (#300)
- Cursor pointer appears on navigation links without a menu (#304)
- Submenu headers do not need to be links to be properly aligned

Still needs tests but if anyone is free we could start a JS/CSS/functionality code review.


stephaniehobson:
Actually, I'll split this into two smaller reviews.

PR #309, opened 2019-02-12
Enable second navigation menu on pages (#300)

stephaniehobson:
- make navigation button code only open its associated menu
- update navigation template macro
  - add download button and mozilla logo as optional parameters
  - remove id from template as it was not being consumed
- add second menu to navigation demo page

### DIFF
--- a/src/assets/js/protocol/protocol-navigation.js
+++ b/src/assets/js/protocol/protocol-navigation.js
@@ -64,8 +64,8 @@ if (typeof Mzp === 'undefined') { // eslint-disable-line block-scoped-var
             for (var i = 0; i < navButtons.length; i++) {
                 navButtons[i].addEventListener('click', Navigation.onClick, false);
             }
+            Navigation.setAria();
         }
-        Navigation.setAria();
     };
 
     /**
--- a/src/assets/js/protocol/protocol-navigation.js
+++ b/src/assets/js/protocol/protocol-navigation.js
@@ -11,7 +11,7 @@ if (typeof Mzp === 'undefined') { // eslint-disable-line block-scoped-var
     'use strict';
 
     var Navigation = {};
-    var navItems;
+    var navItemsLists;
     var _options = {
         onNavOpen: null
     };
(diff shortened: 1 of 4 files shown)


stephaniehobson:
Needs tests but the other code is ready for a first pass.

src/assets/js/protocol/protocol-navigation.js
```diff
@@ -47,18 +49,23 @@ if (typeof Mzp === 'undefined') { // eslint-disable-line block-scoped-var
...
-            document.querySelector('.mzp-c-navigation-menu-button').addEventListener('click', Navigation.onClick, false);
-            Navigation.setAria();
+        navItemsLists = document.querySelectorAll('.mzp-c-navigation-items');
+        if (navItemsLists.length > 0) {
+            var navButtons = document.querySelectorAll('.mzp-c-navigation-menu-button');
+            for (var i = 0; i < navButtons.length; i++) {
+                navButtons[i].addEventListener('click', Navigation.onClick, false);
+            }
         }
+        Navigation.setAria();
```
alexgibson: Could `setAria()` be inside the `if (navItemsLists.length > 0) {}` conditional?

src/patterns/organisms/navigation/navigation.hbs
```diff
@@ -18,11 +18,13 @@ links:
   <div class="mzp-c-navigation-l-content">
     <div class="mzp-c-navigation-container">
       <button class="mzp-c-navigation-menu-button" type="button" aria-controls="map-c-navigation-items">Menu</button>
-      <div class="mzp-c-navigation-logo"><a href="https://www.mozilla.org/">Mozilla</a></div>
+      {{#if logo}}<div class="mzp-c-navigation-logo"><a href="https://www.mozilla.org/">Mozilla</a></div>{{/if}}
       <div class="mzp-c-navigation-items" id="mzp-c-navigation-items">
```
alexgibson: We should probably remove ids like `id="mzp-c-navigation-items"` from the component if we're now using more than one in a single page?

src/assets/js/protocol/protocol-navigation.js
```diff
@@ -20,17 +20,19 @@ if (typeof Mzp === 'undefined') { // eslint-disable-line block-scoped-var
      * Event handler for navigation menu button `click` events.
      */
     Navigation.onClick = function(e) {
+        var thisNavItemList  = e.target.parentNode.querySelector('.mzp-c-navigation-items');
```
alexgibson: If there can now be multiple navigation menus on one page, should we pass `thisNavItemList` as a parameter to `onNavOpen` and `onNavClose`, so those callbacks can work out which menu was interacted with?

stephaniehobson:
I do want the collapse functionality for my use case.

However, I also misunderstood the documentation on the Menu page thinking only the part that displays on hover is the menu. I will add a visual example to that page of the entire menu component since I found it hard to synthesize the code example without it.

src/assets/js/protocol/protocol-navigation.js
stephaniehobson: ✅

src/patterns/organisms/navigation/navigation.hbs
stephaniehobson: ✅ ID is now configurable (we need to keep an ID for the ARIA to work properly)

src/assets/js/protocol/protocol-navigation.js
stephaniehobson: ✅

stephaniehobson:
Updated.

PR #321, opened 2019-02-18
Make the Menu molecule more resilient to JS errors (Fixes #320)

### LINKED ISSUE #320: Make the Menu molecule more resilient to JS errors
Right now, our main navigation works both with JavaScript enabled and disabled which is good. We do this by relying on the `no-js` and `js` classes as style hooks, which get added by the [base script here](https://github.com/mozilla/protocol/blob/master/src/assets/js/protocol/protocol-base.js#L19) in the `<head>`.

We also use a feature detect in the [menu JS](https://github.com/mozilla/protocol/blob/master/src/assets/js/protocol/protocol-menu.js#L332-L337), which adds a fallback CSS class for older browsers that do run JS, but don't support all the features the menu requires.

Between both of these pieces of JS executing however, there is potentially a lot that could happen. A script that's in the same bundle as the menu JS could throw an error for example, halting execution before the menu gets initialized. In such a scenario, the nav could end up in a broken or semi-functional state.

I'd like to try and improve this by changing the menu not to rely on the `no-js` and `js` classes. Instead, we could have a default class to indicate basic functionality, such as `mzp-c-menu mzp-is-basic`. We could then swap this out for something like `mzp-c-menu mzp-is-enhanced` once the menu JS executes. Doing this would make the menu much more resilient to any potential errors that may happen during the course of a page load, since it would no longer rely on an entirely unrelated script.

Visually, I don't think there _should_ be any noticeable difference here when doing it this way, since the menu only shows on hover in either scenario, and on smaller viewports it is hidden behind the hamburger menu.

alexgibson:
## Description

This PR makes the menu molecule more resilient to JS errors, by no longer relying on the `js` and `np-js` global class names.

### Issue

#320

### Testing

- [x] Menu should still work as expected with JS enabled.
- [x] Menu should still work as expected with JS disabled.
- [x] Menu should still function when an unexpected JS error is encountered (I misnamed a function in menu.js locally to recreate the scenario).

### DIFF
--- a/src/assets/js/protocol/protocol-menu.js
+++ b/src/assets/js/protocol/protocol-menu.js
@@ -30,10 +30,10 @@ if (typeof Mzp === 'undefined') { // eslint-disable-line block-scoped-var
      */
     Menu.open = function(el, animate) {
         if (animate) {
-            el.classList.add('is-animated');
+            el.classList.add('mzp-is-animated');
         }
 
-        el.classList.add('is-selected');
+        el.classList.add('mzp-is-selected');
 
         _menuOpen = true; // For checking menu state on keyup.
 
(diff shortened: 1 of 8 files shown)


stephaniehobson:
We lost a line from the changelog in the rebase :(

alexgibson:
> We lost a line from the changelog in the rebase :(

No we didn't, I just moved it down under bug fixes :)

src/assets/js/protocol/protocol-menu.js
```diff
@@ -297,12 +297,15 @@ if (typeof Mzp === 'undefined') { // eslint-disable-line block-scoped-var
...
      */
-    Menu.cssFallback = function() {
-        var menu = document.querySelector('.mzp-c-menu');
-        var currentClassName = menu.className;
-        menu.className = currentClassName.replace(/mzp-c-menu/, 'mzp-c-menu mzp-c-menu-basic');
+    Menu.enhanceJS = function() {
+        var menu = document.querySelectorAll('.mzp-c-menu');
+
+        for (var i = 0; i < menu.length; i++) {
+            menu[i].classList.remove('mzp-is-basic');
```
stephaniehobson: Damn. As a team we've been using `.mzp-is-state` and `.is-state` inconsistently.

The menu component already uses `.is-animated` and `.is-selected`.

I opened https://github.com/mozilla/protocol/issues/324 to discuss.

alexgibson: Given that the docs suggest using a namespace, I'll update here so the menu component is consistent 👍

alexgibson:
Added another commit to update the menu and navigation components to consistently use `mzp-is-*` classes.

PR #323, opened 2019-02-19
Add example menu to menu molecule page (Fix: #322)

### LINKED ISSUE #322: Add example menu to menu molecule page
I was confused about what was included in a menu 😕

Clearly I need to read more better.

But also, we could add a demo to show the difference between a Menu and a Navigation.

stephaniehobson:
## Description

Adds an example of a menu to the menu page in the docs.

### Issue

Fix #322 

### Testing

The menu drop down overflows the sample container. I experimented with containing it inside the container but then it ends up really squished. It's buggy but it accurately represents the menu behaviour too. 🤷‍♀️

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Features
 
+* **docs:** Add example menu to menu molecule page (#322)
 * **css:** Define basic styles for dl lists (#295)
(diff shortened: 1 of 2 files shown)


stephaniehobson:
Icons removed.

PR #325, opened 2019-02-21
Add optional label to navigation mobile menu button. #313

### LINKED ISSUE #313: Add optional label to navigation mobile menu button.
People are more likely to click on hambuger menus when they know what it will do. Simply adding the word "menu" seems to increase clicks.

If there are two menus on a page together (#309) it will be especially important to label one of them.

stephaniehobson:
## Description

Add optional label to navigation mobile menu button.

screen shot 2019-02-20 at 4 38 36 pm

### Issue

Fix #313

### Testing

- RTL
- small and large screens
- label displays when it should, does not when it shouldn't
- no demo for this right now, it makes sense to add it after #309 merges

### DIFF
--- a/src/assets/sass/protocol/components/_navigation.scss
+++ b/src/assets/sass/protocol/components/_navigation.scss
@@ -152,7 +152,7 @@
 .mzp-c-navigation-menu-button {
     @include bidi(((background-position, right 6px center, left 6px center ),));
     @include bidi(((float, right, left),));
-    @include bidi(((padding-right, 32px, padding-left, 0),));
+    @include bidi(((padding, 0 32px 0 6px, 0 6px 0 32px),));
     background-color: transparent;
     background-image: url('#{$image-path}/icons/ui/menu-black.svg');
     background-repeat: no-repeat;
--- a/src/assets/sass/protocol/components/_navigation.scss
+++ b/src/assets/sass/protocol/components/_navigation.scss
@@ -150,16 +150,16 @@
 // Mobile Navigation Icon
 
 .mzp-c-navigation-menu-button {
+    @include bidi(((background-position, right 6px center, left 6px center ),));
     @include bidi(((float, right, left),));
-    @include image-replaced;
-    background: transparent url('#{$image-path}/icons/ui/menu-black.svg') center center no-repeat;
+    @include bidi(((padding-right, 32px, padding-left, 0),));
+    background-color: transparent;
+    background-image: url('#{$image-path}/icons/ui/menu-black.svg');
+    background-repeat: no-repeat;
     border-radius: 4px;
     border: none;
-    cursor: pointer;
     display: none;
     height: 32px;
-    padding: 0;
-    width: 32px;
 
     &:hover,
     &:active,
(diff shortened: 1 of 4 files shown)


alexgibson:
Looks like this one needs a rebase

stephaniehobson:
Rebased.

stephaniehobson:
And added demo.

src/assets/sass/protocol/components/_navigation.scss
```diff
@@ -169,6 +169,15 @@
     }
 }
 
+.mzp-c-navigation-menu-button:not(.has-label) {
```
alexgibson: Not a blocker, but could we improve legacy browser support a bit if we made the default not having a label? (removing the reliance on `:not()`?

src/assets/sass/protocol/components/_navigation.scss
```diff
@@ -150,16 +150,16 @@
 // Mobile Navigation Icon
 
 .mzp-c-navigation-menu-button {
+    @include bidi(((background-position, right 6px center, left 6px center ),));
     @include bidi(((float, right, left),));
-    @include image-replaced;
-    background: transparent url('#{$image-path}/icons/ui/menu-black.svg') center center no-repeat;
+    @include bidi(((padding-right, 32px, padding-left, 0),));
```
alexgibson: Looks like RTL is missing some right margin value

screenshot 2019-02-26 at 09 26 30

src/patterns/organisms/navigation/navigation.hbs
```diff
@@ -17,7 +18,7 @@ links:
 <div class="mzp-c-navigation">
   <div class="mzp-c-navigation-l-content">
     <div class="mzp-c-navigation-container">
-      <button class="mzp-c-navigation-menu-button" type="button" aria-controls="{{ id }}">Menu</button>
+      <button class="mzp-c-navigation-menu-button{{#if button-label}} has-label{{/if}}" type="button" aria-controls="map-c-navigation-items">{{#if button-label}}{{ button-label }}{{ else }}Menu{{/if}}</button>
```
alexgibson: Is it OK to set `aia-controls` to point to a class? I thought it needed an ID?

src/assets/sass/protocol/components/_navigation.scss
stephaniehobson: I ran into some problems over-riding the image replacement when I did it the other way. I can try again but do we even serve IE8 these files?

src/assets/sass/protocol/components/_navigation.scss
stephaniehobson: ✅

src/patterns/organisms/navigation/navigation.hbs
stephaniehobson: Not only is it not okay... I miss spelled it. 🤦‍♀️ 

✅

stephaniehobson:
Updated.

src/assets/sass/protocol/components/_navigation.scss
alexgibson: I don't think IE8 would ever seen the mobile menu button, I was thinking more of the long tail of older mobile devices out there. I took a look on [caniuse](https://caniuse.com/#feat=css-sel3) and support seems pretty broad. I think we're likely fine to use it 👍

PR #349, opened 2019-03-23
Use light-links mixin in t-dark theme #335

### LINKED ISSUE #335: Visited link colour in t-dark themes too dark
It's still purple like the light theme 😢

Screen Shot 2019-03-12 at 2 44 21 PM

stephaniehobson:
## Description

Use light-links mixin in `.t-dark` theme links.
Add visited links styles to light-links mixin.

### Issue

Fix #335 

### Testing

This underlined all child buttons under `.t-dark` themed elements... hence the important. Let me know if you think this is a bad idea I'm a bad person and I should feel bad.

### DIFF
--- a/src/assets/sass/protocol/includes/_mixins.scss
+++ b/src/assets/sass/protocol/includes/_mixins.scss
@@ -234,9 +234,14 @@
         text-decoration: underline;
     }
 
+    a:visited {
+        opacity: 0.8;
+    }
+
     a:hover,
     a:focus,
     a:active {
+        opacity: 1;
         text-decoration: none;
     }
 }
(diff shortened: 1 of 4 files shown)


src/assets/sass/protocol/includes/_mixins.scss
```diff
@@ -234,9 +234,14 @@
         text-decoration: underline;
     }
 
+    a:visited {
+        opacity: 0.8;
```
craigcook: I haven't tested, but might this turn buttons translucent? If not then this is great, but if it does maybe we just need an explicit `opacity: 1;` override in the button atom. Or alternatively we could try `color: transparentize($color-white, .2);`?

stephaniehobson: You're so smart. 🤯

stephaniehobson:
Updated. I had to bump the specificity of the button link styles by adding the `a` element. This feels increasingly messy but solved a bug I found on the [dark hero demos](http://localhost:3000/demos/hero.html).

PR #369, opened 2019-04-08
Fix at2x error and publish

### LINKED ISSUE #368: Error in at2x mixin
`@if $w = 'contain' {` should be  `@if $w == 'contain' {`

stephaniehobson:
## Description

Fixes a bug in at2x mixin and publishes version 6.0.1.

### Issue

Fix #368

### Testing

I found it useful to  copy the mixin into code pen and try the 4 following combinations to make sure the right thing is being output:

- no size attributes, should output auto
- just a width (height should be auto)
- width and height
- just contain (should be no height: auto).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 6.0.1
+
+# Bug Fixes
+
+* **css:** Error in at2x mixin #368
+
 # 6.0.0
 
 # Bug Fixes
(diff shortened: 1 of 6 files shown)


stephaniehobson:
Let me know if I should separate the bug fix from the version bump. I would like to publish in a hurry though to unblock https://github.com/mozilla/bedrock/pull/7048

Issue #297, opened 2019-02-07
Create new atomic heading sizes


alexgibson:
Protocol already has a good set of default heading sizes, but at the component level we override those defaults using component-specific classes:

- `mzp-c-call-out-title`
- `mzp-c-card-feature-title`
- `mzp-c-card-title`
- `mzp-c-hero-title`
- `mzp-c-billboard-title`
- `mzp-c-card-picto-title`
- `mzp-c-article-title`
- `mzp-c-newsletter-title`
- `mzp-c-card-feature-title`

In practical use this has turned out to be too limiting. Designs often require setting different visual heading sizes depending on the hierarchy of the content, and we're also rewriting a lot of the same CSS for each component which is inefficient.

It may be more flexible to have a set of generic heading style classes e.g.

- `mzp-u-title-xs`
- `mzp-u-title-sm`
- `mzp-u-title-md`
- `mzp-u-title-lg`
- `mzp-u-title-xl`

Components could then use these generic classes as needed, and it would also be easy to swap out for a new class if a heading needed a slightly different visual hierarchy.

In addition, headings could also come with some sort of `mzp-has-icon` modifier, which enables incorporating an icon as a pseudo element, as seen in the Call Out and Hero components.

craigcook:
Since these classes are presentational I would suggest a `u-` prefix instead of `c-`.

Alternatively, I've been thinking along the lines of a small set of universal title classes and let the size be determined by their context. Like `mzp-c-title-primary` and `mzp-c-title-secondary` or similar. Basically treating a "title" as a standalone component with some default styling applied to ALL such titles, then the presentation can be refined for different contexts.

Something like:

`.mzp-c-article .mzp-title-primary { @include text-display-xl; }`
`.mzp-c-card .mzp-title-primary { @include text-display-sm; }`

But I'm probably being a semantic purist and I should just accept that presentational utility classes are super useful and pragmatic, and I'll just die a little bit more inside every time I use one.

alexgibson:
https://twitter.com/csswizardry/status/1093083909468889088

stephaniehobson:
Can we safely assume titles should be the appropriate brand font?

stephaniehobson:
McDonald's option for discussion:

https://github.com/mozilla/protocol/pull/387/files

```
mzp-u-title-xxl {
    @include text-display-xxl;

    .mzp-t-mozilla {
        & {
            @include font-mozilla;
        }
    }

    .mzp-t-firefox {
        & {
[...]

PR #387, opened 2019-06-14
Atomic heading utility classes (#297)

stephaniehobson:
## Description

Add atomic heading utility classes.

### Issue

See #297

### DIFF
--- a/src/assets/sass/protocol/base/utilities/_typography.scss
+++ b/src/assets/sass/protocol/base/utilities/_typography.scss
@@ -1,0 +1,118 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import '../../includes/lib';
+
+.mzp-u-title-xxl {
+    @include text-display-xxl;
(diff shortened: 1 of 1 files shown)


craigcook:
OK, I'm gonna pick a fight...

While I acknowledge the pragmatic usefulness of presentational utility classes, I also don't think we should encourage them. I worry about baking a lot of presentation into the HTML by relying on a class that does nothing but set a font size. I'd rather treat titles as truly atomic components, like what KIND of title is this, not just what size it should be. 

Of course, as long as we're getting full-page designs with text sized arbitrarily, we might have to resort to this anyway. But I'm prepared to take that fight to the brand team and work on defining reusable title atoms. 🥊

stephaniehobson:
I think this PR is still good as-is. One outstanding (but not blocking) [question from #297](https://github.com/mozilla/protocol/issues/297#issuecomment-501931689): 

> Now that I'm looking at it I really want to alias @mixin text-display-lg to @mixin text-title-lg. Would that increase or decrease confusion and muscle memory?

stephaniehobson:
Oh, actually this needs documentation still 😅

stephaniehobson:
Closing in favour of https://github.com/mozilla/protocol/pull/555.

Issue #366, opened 2019-04-06
Blockquote seems kind of big


stephaniehobson:
In bedrock the default text size has been decreased to `@include text-display-sm;` and the top and bottom margins increased.

The smaller font size should probably be made default. Maybe also the default margins should be increased.

PR #388, opened 2019-06-25
Add citation style to blockquote and decrease font size. (#366)

amychurchwell:
## Description

- Add citation styles to blockquotes. 
- Decreases font size of blockquotes.

### Issue
#366

### DIFF
--- a/src/assets/sass/protocol/base/elements/_quotes.scss
+++ b/src/assets/sass/protocol/base/elements/_quotes.scss
@@ -9,7 +9,7 @@ blockquote {
         (border-left, 5px solid $color-gray-20, border-right, 0),
         ));
     @include text-display-sm;
-    margin: $spacing-lg auto $spacing-lg;
+    margin: $spacing-lg auto;
     padding: $spacing-sm $spacing-lg;
     font-weight: bold;
 
@@ -36,9 +36,4 @@ blockquote {
     .mzp-t-mozilla & {
         @include font-mozilla;
     }
-}
-
-[dir='rtl'] {
-    cite:before { content: ''; }
-    cite:after { content: ' —'; }
 }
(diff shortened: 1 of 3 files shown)


src/assets/sass/protocol/base/elements/_quotes.scss
```diff
@@ -4,16 +4,24 @@
...
 
-
 blockquote {
     @include bidi((
         (border-left, 5px solid $color-gray-20, border-right, 0),
         ));
-    @include text-display-md;
-    margin: $spacing-md auto;
+    @include text-display-sm;
+    margin: $spacing-lg auto $spacing-lg;
```
stephaniehobson: When using the short hand like this we don't need to specify top and bottom separately if they're the same values.

When there's two values like this it reads as margin: top-and-bottom  left-and-right;

I think you know this though, since you didn't repeat the left-and-right values. So maybe this is left over from some other tweaking.

amychurchwell: Whoops! You're right.  👍

src/assets/sass/protocol/base/elements/_quotes.scss
```diff
@@ -29,3 +37,8 @@ blockquote {
         @include font-mozilla;
     }
 }
+
+[dir='rtl'] {
```
stephaniehobson: Testing the PR, I don't think you need this. The expected behaviour is to have the line on the right hand side of the citation for RTL languages, and that is where it appears for me without this.

src/patterns/atoms/typographic/quotes.hbs
```diff
@@ -11,10 +11,12 @@ order: 5
   <p>
     "{{data "specimens.sentence"}}"
   </p>
+  <cite>A. Nonymous</cite>
```
stephaniehobson: 🔥

stephaniehobson:
@amychurchwell If you give this a rebase the tests should pass now thanks to @alexgibson.

amychurchwell:
@stephaniehobson Ok, gave it a rebase & it seems to have fixed the tests. :)

Issue #178, opened 2018-07-17
Write unit tests for protocol-modal.js


alexgibson:
Code was ported from bedrock, so will likely need some refactoring to make it easier to write tests for.

craigcook:
May get bundled in with https://github.com/mozilla/protocol/issues/255

PR #431, opened 2019-07-18
Add notification bar component. (Fixes #383)

### LINKED ISSUE #383: Create notification bar component
The upcoming Trailhead designs show a new notification both on /thanks and on /whatsnew

image

It also appears in our new brand guidelines:

Screenshot 2019-05-10 at 10 43 17

Things to note:

- There are a range of new colors here that are not currently in our tokens.
- The bar seems like it needs to support an optional button to close the notification.

---
Estimates |  |
--- | --- |
Eng | 5 |

amychurchwell:
## Description

Adds the notification bar component used on /thanks and /whatsnew

### Issue

#383
https://github.com/mozilla/bedrock/issues/7468

### Testing
https://demo2--mozilla-protocol.netlify.com/patterns/molecules/notification-bar.html

- Supports different styles:
  - Success
  - Warning
  - Error
  - Click

- Dismiss is optional.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -186,6 +186,7 @@
       "version": "0.1.4",
       "resolved": "https://registry.npmjs.org/align-text/-/align-text-0.1.4.tgz",
       "integrity": "sha1-DNkKVhCT810KmSVsIrcGlDP60Rc=",
+      "optional": true,
       "requires": {
         "kind-of": "^3.0.2",
         "longest": "^1.0.1",
@@ -3263,7 +3264,8 @@
         },
         "ansi-regex": {
           "version": "2.1.1",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "aproba": {
           "version": "1.2.0",
@@ -3281,11 +3283,13 @@
         },
         "balanced-match": {
           "version": "1.0.0",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
+          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
@@ -3298,15 +3302,18 @@
         },
         "code-point-at": {
           "version": "1.1.0",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "concat-map": {
           "version": "0.0.1",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "console-control-strings": {
           "version": "1.1.0",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "core-util-is": {
           "version": "1.0.2",
@@ -3409,7 +3416,8 @@
         },
         "inherits": {
           "version": "2.0.3",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "ini": {
           "version": "1.3.5",
@@ -3419,6 +3427,7 @@
         "is-fullwidth-code-point": {
           "version": "1.0.0",
           "bundled": true,
+          "optional": true,
           "requires": {
             "number-is-nan": "^1.0.0"
           }
@@ -3431,17 +3440,20 @@
         "minimatch": {
           "version": "3.0.4",
           "bundled": true,
+          "optional": true,
           "requires": {
             "brace-expansion": "^1.1.7"
           }
         },
         "minimist": {
           "version": "0.0.8",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "minipass": {
           "version": "2.2.4",
           "bundled": true,
+          "optional": true,
           "requires": {
             "safe-buffer": "^5.1.1",
             "yallist": "^3.0.0"
@@ -3458,6 +3470,7 @@
         "mkdirp": {
           "version": "0.5.1",
           "bundled": true,
+          "optional": true,
           "requires": {
             "minimist": "0.0.8"
           }
@@ -3530,7 +3543,8 @@
         },
         "number-is-nan": {
           "version": "1.0.1",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "object-assign": {
           "version": "4.1.1",
@@ -3540,6 +3554,7 @@
         "once": {
           "version": "1.4.0",
           "bundled": true,
+          "optional": true,
           "requires": {
             "wrappy": "1"
           }
@@ -3615,7 +3630,8 @@
         },
         "safe-buffer": {
           "version": "5.1.1",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "safer-buffer": {
           "version": "2.1.2",
@@ -3645,6 +3661,7 @@
         "string-width": {
           "version": "1.0.2",
           "bundled": true,
+          "optional": true,
           "requires": {
             "code-point-at": "^1.0.0",
             "is-fullwidth-code-point": "^1.0.0",
@@ -3662,6 +3679,7 @@
         "strip-ansi": {
           "version": "3.0.1",
           "bundled": true,
+          "optional": true,
           "requires": {
             "ansi-regex": "^2.0.0"
           }
@@ -3700,11 +3718,13 @@
         },
         "wrappy": {
           "version": "1.0.2",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "yallist": {
           "version": "3.0.2",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         }
       }
     },
@@ -5983,7 +6003,8 @@
     "longest": {
       "version": "1.0.1",
       "resolved": "https://registry.npmjs.org/longest/-/longest-1.0.1.tgz",
-      "integrity": "sha1-MKCy2jj3N3DoKUoNIuZiXtd9AJc="
+      "integrity": "sha1-MKCy2jj3N3DoKUoNIuZiXtd9AJc=",
+      "optional": true
     },
     "longest-streak": {
       "version": "2.0.2",
--- a/src/assets/js/protocol/protocol-notification-bar.js
+++ b/src/assets/js/protocol/protocol-notification-bar.js
@@ -45,13 +45,13 @@ Mzp.Notification = (function() { // eslint-disable-line block-scoped-var
 
         // Notification Title
         if (options && options.title){
+
             var notificationTitle = document.createElement('p');
-            notificationTitle.className = 'mzp-c-notification-bar-content';
             notificationTitle.appendChild(title);
 
             // add title to notification
             notification.appendChild(notificationTitle);
-        }else{
+        } else {
             console.log('Notification banner requires title parameter.'); // eslint-disable-line no-console
         }
 
@@ -64,7 +64,7 @@ Mzp.Notification = (function() { // eslint-disable-line block-scoped-var
 
         // Show button & add event listener
         if (options && options.hasDismiss){
-            notificationFragment.childNodes[0].insertAdjacentHTML('beforeend', dismissButton);
+            notificationFragment.childNodes[0].insertAdjacentHTML('afterbegin', dismissButton);
 
             var button = notificationFragment.querySelector('.mzp-c-notification-bar-button');
 
--- a/src/assets/js/protocol/protocol-notification-bar.js
+++ b/src/assets/js/protocol/protocol-notification-bar.js
@@ -44,6 +44,7 @@ Mzp.Notification = (function() { // eslint-disable-line block-scoped-var
         notification.className = 'mzp-c-notification-bar ' + className + ' ' + isSticky;
 
         // Notification Title
+
         if (options && options.title){
 
             var notificationTitle = document.createElement('p');
@@ -51,8 +52,6 @@ Mzp.Notification = (function() { // eslint-disable-line block-scoped-var
 
             // add title to notification
             notification.appendChild(notificationTitle);
-        } else {
-            console.log('Notification banner requires title parameter.'); // eslint-disable-line no-console
         }
 
         // Notification Fragment
--- a/src/assets/js/protocol/protocol-notification-bar.js
+++ b/src/assets/js/protocol/protocol-notification-bar.js
@@ -1,0 +1,93 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// create namespace
+if (typeof Mzp === 'undefined') { // eslint-disable-line block-scoped-var
+    var Mzp = {};
+}
+
+Mzp.Notification = (function() { // eslint-disable-line block-scoped-var
+    'use strict';
+
+    var options = {};
+
+    /*
+    options: object of optional params:
+        title: title to display inside the notification.
+        className: optional CSS class name to apply to the notification.
+        closeText: 'string to use for close button a11y.
+        hasDismiss: boolean - include or not include dismiss button.
+        isSticky: boolean - determines if notification is absolutely positioned and sticky.
+    */
+
+    var _init = function(origin, opts) {
+        options = opts;
+
+        // Create new notification
+        var title = (options && options.title) ? options.title : '';
+        var className = (options && options.className) ? options.className : '';
+        var closeText = (options && options.closeText) ? options.closeText : '';
+        var isSticky = (options && options.isSticky) ? 'mzp-is-sticky' : '';
+
+        // Notification Fragment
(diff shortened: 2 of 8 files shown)


amychurchwell:
Should we support a standard set cookie for components that utilize a dismiss function like this in protocol? I'm not sure if this should be baked in on a per-component basis or a reusable token(?) of some kind.

alexgibson:
My personal take here is the Protocol shouldn’t use or require setting any form of cookies in its code (that’s probably a bigger discussion than that we’ve yet to have). Providing a callback when the notification is dismissed should be enough so that consumers (i.e bedrock) can then use cookies if required.

It might be nice to have a JavaScript API that enables showing the notification itself too, vs a script that does it automatically on load? Also, tests!

craigcook:
A few drive-by opinions, not a review because I know it's WIP.

It probably shouldn't be an `h1`. The original one on the whatsnew page was because it's arguably the most important content on that particular page, but as a general-purpose notification the example should be a plain old `p`, or maybe a `div` or `aside` with a `p` for the content.

We should move the close button to the left in RTL.

And this is more of a discussion for the designers, but I'm not sure the "click" variant should even exist. It appears in the brand guide but I think it was just an example of "use blue for actions," it wasn't a fully thought-through component. If there's an action to be taken it shouldn't take the form of a big pseudo-button with its own little dismiss button inside it. We should get some more opinions on that.

I also think we might need a neutral notification for messages that aren't warnings, errors, or successes, just general notices. Maybe with a light gray background, or white with a blue or purple border?

amychurchwell:
@stephaniehobson If you have a moment tomorrow, would you mind taking a look at this? I've refactored the JS a bit but I'm struggling with the unit test portion (or perhaps the JS needs to be refactored more?) 😔

amychurchwell:
@alexgibson would you mind taking a look at this tomorrow? (only if you have the time, is not a p1)
The JS isn't quite right and writing unit tests has been a little blocked.

src/patterns/molecules/notification-bar/notification-bar.hbs
```diff
@@ -0,0 +1,90 @@
...
+        e.currentTarget.parentNode.remove();
+      });
+    }
+
+    var notificationButton = document.querySelector('.mzp-c-button');
+
+    notificationButton.addEventListener('click', function (e) {
+      e.preventDefault();
+
+      Mzp.Notification.init(e.target, content, {
```
alexgibson: The demo is throwing an error when I click the button saying `ReferenceError: content is not defined`.

Looking at the code, `content` does seem to be undefined here, but also the JS does nothing with this value, so is it needed? If I remove this it seems to work ok.

I also notice you're passing `e.target` as the origin here, but you're not using that in the code either. It might be useful to return focus to the origin element when the notification is closed though? :)

tests/unit/protocol-notification-bar.js
```diff
@@ -0,0 +1,42 @@
...
+
+        spyOn(options, 'open');
+        spyOn(options, 'close');
+
+        var button = document.querySelector('.mzp-c-button');
+
+        button.addEventListener('click', function (e) {
+            e.preventDefault();
+
+            Mzp.Notification.init( {
```
alexgibson: The test is failing because the first parameter of the function should be the origin (`button`), and the second parameter should be the `options` (assumign you remove the redundant `content` parameter in the above comment).

I'd also suggest moving this code inside the `it()` function below. If there are to be multiple tests, and this is code that should be run before each, then it should go inside a `beforeEach()` function.

src/assets/js/protocol/protocol-notification-bar.js
```diff
@@ -0,0 +1,85 @@
...
+    /*
+    options: object of optional params:
+        title: title to display inside the notification.
+        className: optional CSS class name to apply to the notification.
+        closeText: 'string to use for close button a11y.
+        hasDismiss: boolean - include or not include dismiss button.
+        isSticky: boolean - determines if notification is absolutely positioned and sticky.
+    */
+
+    var _init = function(origin, content, opts) {
```
alexgibson: Note: `origin` and `content` here is what's causing the confusion :)

amychurchwell:
@alexgibson thanks for the feedback, Alex! I tidied things up and everything seems to be working as it should. I only have two unit tests for opening & closing a notification, I wasn't sure what else to include?
Also, in it's current form, does this component feel ready to use on whatsnew pages? (I'm not quite sure if the way I've written the init of this component _requires_ the use of a "trigger", whereas on wnp the bar is static.)

src/assets/sass/protocol/components/_notification-bar.scss
```diff
@@ -0,0 +1,101 @@
...
+        padding: 0;
+        width: 42px;
+
+        &:hover,
+        &:focus {
+            cursor: pointer;
+        }
+
+        &:focus {
+            outline: 1px dotted $color-white;
```
alexgibson: This focus state is pretty low contrast and hard to see compared to browser defaults?

src/patterns/molecules/notification-bar/notification-bar.hbs
```diff
@@ -0,0 +1,90 @@
...
+notes: |
+    - Notification bars have a range of different colors
+    - `mzp-t-success`
+    - `mzp-t-warning`
+    - `mzp-t-error`
+    - `mzp-t-click`
+    - An optional dismiss button
+    - `mzp-has-dismiss`
+    - And optional stickiness
+    - `mzp-is-sticky`
```
alexgibson: Suggest adding documentation here as to what options the JavaScript API comes with?

src/assets/js/protocol/protocol-notification-bar.js
```diff
@@ -0,0 +1,93 @@
...
+
+    var options = {};
+
+    /*
+    options: object of optional params:
+        title: title to display inside the notification.
+        className: optional CSS class name to apply to the notification.
+        closeText: 'string to use for close button a11y.
+        hasDismiss: boolean - include or not include dismiss button.
+        isSticky: boolean - determines if notification is absolutely positioned and sticky.
```
alexgibson: Missing open and close callback here?

src/assets/js/protocol/protocol-notification-bar.js
```diff
@@ -0,0 +1,93 @@
...
+
+        // Add notification to page
+        document.body.insertBefore(notificationFragment, document.body.childNodes[0]);
+
+        // Remember which element opened the notification for later focus
+        origin.classList.add('mzp-c-notification-origin');
+
+
+        // Execute (optional) open callback
+        if (options && typeof(options.onCreate) === 'function') {
```
alexgibson: Suggest calling this `onNotificationOpen` to be more consistent with other components?

src/assets/js/protocol/protocol-notification-bar.js
```diff
@@ -0,0 +1,93 @@
...
+        }
+    };
+
+    var _closeNotification = function(e) {
+
+        // remove notification from the page
+        e.currentTarget.parentNode.remove();
+
+        // execute (optional) callback
+        if (options && typeof(options.onDestroy) === 'function') {
```
alexgibson: Suggest calling this `onNotificationClose` to be more consistent with other components?

src/assets/js/protocol/protocol-notification-bar.js
```diff
@@ -0,0 +1,93 @@
...
+            options.onDestroy();
```
alexgibson: Would it also be worth passing `e.target` to the close method, so a callback can work out which notification was closed? (should there every be more than one).

src/patterns/molecules/notification-bar/notification-bar.hbs
```diff
@@ -0,0 +1,90 @@
...
+
+<script>
+  (function () {
+    'use strict';
+
+    var dissmissButtons = document.querySelectorAll('.mzp-c-notification-bar-button');
+    for (var i = 0; i < dissmissButtons.length; i++) {
+      dissmissButtons[i].addEventListener('click', function(e) {
+        e.currentTarget.parentNode.remove();
+      });
```
alexgibson: Nit: there are a couple of places in this PR where you're omitting `false` for useCapture on addEventListener. Include it for consistency?

src/patterns/molecules/notification-bar/notification-bar.hbs
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    var notificationButton = document.querySelector('.mzp-c-button');
```
alexgibson: Suggest adding a slightly more specific selector to target this (what if we add a button somewhere else on the page in the future?) :)

alexgibson:
Styling on smaller viewports needs a little work:

image

src/assets/sass/protocol/components/_notification-bar.scss
```diff
@@ -0,0 +1,101 @@
...
+@import "../includes/lib";
+
+.mzp-c-notification-bar {
+    @include text-display-xxs;
+    background-color: $color-gray-20;
+    border-radius: 4px;
+    border: 1px;
+    box-shadow: 1px 1px 1px rgba(27,31,35,.1);
+    margin: $layout-xs auto 0;
+    height: 42px;
```
alexgibson: This fixed height might be troublesome for smaller screens and locales that have long string.

alexgibson:
> Also, in it's current form, does this component feel ready to use on whatsnew pages? (I'm not quite sure if the way I've written the init of this component requires the use of a "trigger", whereas on wnp the bar is static.)

I think the way you've written it feels right, but we should also consider that on some pages we may not require the use of JavaScript at all. For example, this bar is already on [/firefox/download/thanks/](Also, in it's current form, does this component feel ready to use on whatsnew pages? (I'm not quite sure if the way I've written the init of this component requires the use of a "trigger", whereas on wnp the bar is static.)) (minus the close button). As long as we have the option to include it statically on pages without needing JavaScript, then what you're added here is a nice progressive enhancement, should we ever need it 👍

alexgibson:
Oh, the bar on /download/thanks/ also includes a link:

image

I'm guessing this should also be factored into the component design?

package-lock.json
```diff
@@ -186,6 +186,7 @@
       "version": "0.1.4",
       "resolved": "https://registry.npmjs.org/align-text/-/align-text-0.1.4.tgz",
       "integrity": "sha1-DNkKVhCT810KmSVsIrcGlDP60Rc=",
+      "optional": true,
```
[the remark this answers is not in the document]
amychurchwell: I have npm `6.9.0`. These changes occur when I run `npm start`, should I revert or does something need updating?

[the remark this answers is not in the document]
alexgibson: You have the latter version, so don’t worry it’s me who’s out of date 🙂

src/assets/js/protocol/protocol-notification-bar.js
amychurchwell: Happy to make this change! I was following the format the modal component uses:
```
onCreate: function to fire after modal has been created.
onDestroy: function to fire after modal has been closed.
```

src/assets/js/protocol/protocol-notification-bar.js
```diff
@@ -0,0 +1,93 @@
...
+                _closeNotification(e);
+            });
+        }
+
+
+        // Add notification to page
+        document.body.insertBefore(notificationFragment, document.body.childNodes[0]);
+
+        // Remember which element opened the notification for later focus
+        origin.classList.add('mzp-c-notification-origin');
```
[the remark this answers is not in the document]
amychurchwell: that makes a lot of sense! I will definitely make that change. I was following the style in `protocol-modal.js` should we make an issue to update there as well?:
https://github.com/mozilla/protocol/blob/1d08e5a228512ecd95e69be9b664f7ce18ff0cd1/src/assets/js/protocol/protocol-modal.js#L99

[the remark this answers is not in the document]
alexgibson: The modal Js was prettty much copy / pasted from bedrock right in the early days of Protocol. It really needs refactoring, and doesn't even have unit tests. I would save this for https://github.com/mozilla/protocol/issues/178

src/assets/js/protocol/protocol-notification-bar.js
alexgibson: The modal JS was copy / pasted from bedrock in the early days and really needs refactoring. I would probably follow the example of newer components in the library.

amychurchwell: Oh ok, that makes a lot of sense. 🙂

src/assets/js/protocol/protocol-notification-bar.js
```diff
@@ -0,0 +1,112 @@
...
+
+        // Notification Title
+        if (options && options.title){
+            var notificationTitle = document.createElement('p');
+            notificationTitle.className = 'mzp-c-notification-bar-content';
+            notificationTitle.appendChild(title);
+
+            // add title to notification
+            notification.appendChild(notificationTitle);
+        }else{
```
alexgibson: NIt: spaces before / after braces

src/assets/js/protocol/protocol-notification-bar.js
```diff
@@ -0,0 +1,112 @@
...
+            console.log('Notification banner requires title parameter.'); // eslint-disable-line no-console
```
alexgibson: I don't think we should leave `console.log()` in as production code, since it can cause errors in older browsers unless dev tools are open.

tests/unit/protocol-notification-bar.js
```diff
@@ -0,0 +1,54 @@
...
+
+                Mzp.Notification.init(button, {
+                    hasDismiss: true,
+                    onNotificationOpen: options.open,
+                    onNotificationClose: options.close
+                });
+            });
+        });
+
+        it('opens as expected', function() {
```
alexgibson: Should this also test that the notification has been added to the DOM?

tests/unit/protocol-notification-bar.js
```diff
@@ -0,0 +1,54 @@
...
+            var button = document.querySelector('.mzp-c-button');
+            button.click();
+            expect(options.open).toHaveBeenCalled();
+        });
+
+        it('closes as expected', function() {
```
alexgibson: Should this also test that the notification has been removed from the DOM?

src/assets/sass/protocol/components/_notification-bar.scss
```diff
@@ -0,0 +1,112 @@
...
+        height: 20px;
+        padding: 10px;
+
+        &:hover,
+        &:focus {
+            cursor: pointer;
+        }
+
+        @media #{$mq-sm} {
+            @include image-replaced;
```
alexgibson: On breakpoints smaller than this I can see the close button text overflowing the bar, because this only applies to the `mq-sm` breakpoint and above.

The close button at this screen size also seems to be pretty small, which might be difficult as a touch target. I'd suggest maybe making some of these properties the default style?

image

src/assets/js/protocol/protocol-notification-bar.js
amychurchwell: agreed! will fix that up tomorrow.

alexgibson:
One other thought: we should consider a11y when a notification is programatically opened using JS. What should happen: should focus move to the notification title when it is displayed? Should that be required, or optional? Do we need any additional ARIA information?

amychurchwell:
Tests are passing locally, I'm not sure why travis is failing. Also, I'm not entirely sure what additional ARIA information may be useful here. @alexgibson

tests/unit/protocol-notification-bar.js
```diff
@@ -0,0 +1,57 @@
...
+        });
+
+        it('opens as expected', function() {
+            expect(options.open).toHaveBeenCalled();
+
+            var notification = document.querySelector('.mzp-c-notification-bar');
+            expect(notification).toBeTruthy();
+        });
+
+        it('closes as expected', function() {
```
alexgibson: Here you've created an ordered dependency on the "open" test needing to run before the "close" test, which will lead to intermittent failures. Tests run in a random order and not sequentially (hence why this passes sometimes and not at other times, like in CI). 

I'd suggest treating each test as it's own distinct piece of logic. Before each test runs you should create the notification, and after each test finishes you should remove the notification. That way, tests should pass regardless of the order they are run.

tests/unit/protocol-notification-bar.js
```diff
@@ -0,0 +1,64 @@
+/* global describe, beforeEach, it, spyOn */
+
+describe('protocol-notification.js', function() {
+
+    'use strict';
+
+    beforeEach(function () {
```
alexgibson: Nit: Suggest adding an `afterEach` below here to remove the button after each test also?

amychurchwell: Ah! Yes! Good catch!

Issue #385, opened 2019-05-23
Add emphasis box component


stephaniehobson:
This box with shadow and padding is showing up in designs enough that we should make it a 
component:

<img width="720" alt="Screen Shot 2019-05-23 at 1 08 50 PM" src="https://user-images.githubusercontent.com/854701/58283704-d2963280-7d5d-11e9-8db6-aed8845a7555.png">

<img width="724" alt="Screen Shot 2019-05-23 at 1 22 49 PM" src="https://user-images.githubusercontent.com/854701/58283715-e0e44e80-7d5d-11e9-8e43-809e716fd25e.png">

stephaniehobson:
See also mozilla/protocol-tokens#51

johngruen:
is there a measurement spec for this or just a static image?

amychurchwell:
@johngruen design linked in bedrock issue for page containing this component, also static image- https://github.com/mozilla/bedrock/issues/7354#issuecomment-511815158 Should we track down sketch file? I personally feel that we have solid direction for the dev work here.

PR #441, opened 2019-07-22
Emphasis box [Fixes:  #385]

Errolyn:
## Description

This adds an emphasis box to the design system. 

### Issue

 issue: #385


Errolyn:
I think I have figured out how to make Travis happy, hopefully the package-lock update will fix it.

src/assets/sass/protocol/components/_emphasis-box.scss
```diff
@@ -0,0 +1,47 @@
...
+
+.mzp-c-box-emphasis-fuzzy {
+    background: $color-white;
+    box-shadow: 0 0 15px 0 $color-gray-30;
+    border-radius: 6px;
+}
+
+.mzp-c-box-emphasis {
+    background: $color-white;
+    box-shadow: 0 0 10px 0 $color-gray-30;
```
craigcook: We should use a transparent shadow so it will also work on colored backgrounds. This could either be `rgba(0, 0, 0, .25)` (or something like that, not sure of the correct alpha value) or you can use a Sass function like `transparentize($color-black, .75)`.

Errolyn:
I incorporated the feed back and removed the changes to the package.json file. Should be ready for another review.

PR #466, opened 2019-08-20
Publish 8.0.0

stephaniehobson:
## Description

Update to 8.0.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
-# HEAD
+# 8.0.0
 
 ## Features
 
 * **css:** Add Notification bar component (#383)
 * **css:** Details button gets cursor:pointer when hovered (#367)
-* **css:** Add Metropolis as Firefox brand font (#386)
+* **css:** (breaking) Add Metropolis as Firefox brand font (#386)
 
 ## Bug Fixes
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,10 +1,10 @@
-# HEAD
+# 7.1.0
 
 ## Features
 
 * **css:** Add Notification bar component (#383)
 * **css:** Details button gets cursor:pointer when hovered (#367)
-* **css:** Add Metropolis as Firefox brand font (#386)
+* **css:** (breaking) Add Metropolis as Firefox brand font (#386)
```
alexgibson: If we're calling something out as breaking then we should be doing a major version bump. If we think this is safe, then I think we can remove `(breaking)`.

docs/README.md
```diff
@@ -508,6 +508,7 @@ Protocol is published to NPM under the `@mozilla-protocol/core` namespace/packag
 1. Before you start make sure the project's [CHANGELOG.md][changelog] is up to date.
 2. Update the package `version` number in [src/assets/package/package.json][package.json] (use [Semantic Versioning][semver] to determine what the new version number should be).
 3. Update the package README [assets/package/README.md][readme].
+4. Run `npm install` to update the package-lock.json file.
```
alexgibson: Technically this shouldn't be required, since running `npm test` will kick off a bunch of pre execution scripts that include dependency installation. That said, I don't think a little redundancy hurts to err on the safe side 👍

docs/README.md
```diff
@@ -508,6 +508,7 @@ Protocol is published to NPM under the `@mozilla-protocol/core` namespace/packag
...
 4. Submit a pull request with your changes (or commit directly to `master` if you have permission). Once the changes have been merged to master:
```
alexgibson: This should be incremented to `5.` etc.

PR #469, opened 2019-08-21
Narrow content [Fixes #232]

### LINKED ISSUE #232: Add an option for narrower content
Our standard content width is set by the layout class `.mzp-l-content` and it adjusts to the window at different breakpoints, with a maximum width of 1420px (the `$content-max` token). But some text-heavy pages should have a narrower maximum width, especially to keep line lengths comfortable. 

We should add a new layout class or possibly a utility class to define a narrower content column. Maybe just reset the max-width to `$content-md` (736px) or `$content-lg` (992px) with some allowances for padding.

Errolyn:
## Description

adds a smaller content layout and example pages to protocol.

### Issue

https://github.com/mozilla/protocol/issues/232


src/assets/sass/protocol/base/elements/_sections.scss
```diff
@@ -23,6 +23,28 @@
     }
 }
 
+.mzp-s-content {
```
stephaniehobson: Sadly the `l` in `mzp-l-content` stands for "layout" not "large". Though, now I kind of wish it did. Here's [an explanation for the namespaces](https://protocol.mozilla.org/docs/naming.html).

stephaniehobson: Rather than duplicating the entire declaration I think it would be more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) to make the narrow content a theme variation of the existing `mzp-l-content` class.

If people declared it as `<div class="mzp-l-content mzp-t-narrow">`. The only thing you would have to over-ride is the `max-width` declared on line 12 :)

src/patterns/templates/content-layout/small-content-layout.hbs
```diff
@@ -0,0 +1,17 @@
+---
+name: Content layouts
```
stephaniehobson: Documentation FTW.

Errolyn: 💃

Errolyn:
I made changes so this should be good to go!

src/assets/sass/protocol/base/elements/_sections.scss
```diff
@@ -23,25 +23,18 @@
...
-    margin: 0 auto;
+.mzp-t-narrow {
     max-width: $content-max;
     min-width: $content-xs;
-    padding: $layout-xs;
-    position: relative;
 
     @media #{$mq-md} {
         max-width: $content-md;
         min-width: $content-xs;
```
stephaniehobson: We don't need to declare the min-width at each break point. They each inherit the definition from the last. Cascade FTW.

PR #473, opened 2019-08-23
Add inline list component (#465)

### LINKED ISSUE #456: Add tokens for box-shadows
There are 5 attributes that contribute a shadow element
- inset (boolean)
- color
- blur
- x
- y
- spread

**Expected CSS output**
```
$shadow-full: box-shadow: 0 8px 24px rgba(0, 0, 0, .1);
$shadow-half: box-shadow: 0 4px 12px rgba(0, 0, 0, .1);
```

**Expected JSON output (for Sketch consumption)**
```
{
    "inset": false,
    "color": '#0000001A',
    "blur":  24,
    "x": 0,
    "y": 8,
    "spread": 0
},
{
    "inset": false,
    "color": '#0000001A',
    "blur":  12,
    "x": 0,
    "y": 4,
    "spread": 0
}
```

### LINKED ISSUE #465: New Component: Inline List
Here is the spec:

list-hor

Going to be used on the new /firefox, browsers and products pages.

Probably best as a utility class.

stephaniehobson:
## Description

Add inline list component

### Issue

Fix #465

### Testing

I added it to the http://localhost:3000/patterns/atoms/lists.html page. Test in RTL, test that list items don't get broken up over multiple lines.


src/assets/sass/protocol/utilities/_inline-list.scss
```diff
@@ -0,0 +1,21 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import '../includes/lib';
+
+.mzp-u-inline-list {
+    margin: $layout-md 0 0 ($layout-md * -1);
+
+    li {
+        display: inline-block;
```
craigcook: Should this be wrapped in a media query so it defaults to stacked on small screens?

stephaniehobson:
@craigcook ready for re-review.

src/assets/sass/protocol/utilities/_inline-list.scss
stephaniehobson: I tested it both ways and as long as the links are short I think it works better like this, even on small screens. If it proves to be a problem once we have actual content we can revisit.

PR #488, opened 2019-09-18
Use window.Mzp in Modal and Notification components

liamnewmarch:
In [github.com/mdn/developer-portal](https://github.com/mdn/developer-portal) we are using Webpack to include Protocol components in our bundle.

In this setup, each component is wrapped in a function which means we can't rely on the `var Mzp` check at the top of each file to create an implicit global – in our case the variable that is created is scoped to the module. This doesn’t matter for most components because assign something to `window.Mzp`, but __Modal__ and __Notification Bar__ use the module-scoped `Mzp` variable instead.

This PR changes that assignment, so __Modal__ and __Notification Bar__ use `window.Mzp`, as well as other small changes for consistency with the rest of the components.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## HEAD
+## 9.0.0
 
 * **assets:** (breaking) Update @mozilla-protocol/assets to 3.0.0 (#479)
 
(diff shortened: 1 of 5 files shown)


liamnewmarch:
Bump!

Is anyone able to review this? If so, it would be helpful in unblocking something we’re working on.

liamnewmarch:
Pushing new commits isn’t working here so I’ve opened #496 instead, sorry about that!

PR #527, opened 2019-11-27
Update buttons to inline-block. (Fixes #486)

### LINKED ISSUE #486: Download buttons should be inline-block instead of block level?
## Description

Download button containers are block level, meaning they take up 100% of their parent's container width. They should probably be inline-block?

### Steps to reproduce

1. Visit https://protocol.mozilla.org/patterns/molecules/download-button.html

### Expected result

image

### Actual result

image

### Environment

Any browser.

tikhsuP:
#486 Fixed.


craigcook:
Oops, my apologies. This same issue was addressed in another pull request (https://github.com/mozilla/protocol/pull/507). I'm going to close this one in favor of that earlier PR.

PR #553, opened 2020-03-11
Notification bar updates

### LINKED ISSUE #549: Notification bar should use box-sizing: border-box;
## Description

Because the notification bar component has [default padding](https://github.com/mozilla/protocol/blob/master/src/assets/sass/protocol/components/_notification-bar.scss#L19) values, the notification container can break out of the viewport at sizes that specify a [min-width](https://github.com/mozilla/protocol/blob/master/src/assets/sass/protocol/components/_notification-bar.scss#L22).

The component should use `@include border-box;` to take into account the extra padding.

### Steps to reproduce

This is much clearer on bedrock where we have the notification is use, but you can see the issue on this URL: https://protocol.mozilla.org/patterns/molecules/notification-bar.html

Resize the browser and notice the notification breaks out of its container at around 520px.

### Expected result

The notification should adapt well to any screen size.

### Actual result

The notification can be wider than it's container.

### Environment

Firefox Nightly

craigcook:
### Description

Fixes #549 - use border-box
Fix image replacement bug
Add focus styles
Tweak spacing

### Issue

#549

### DIFF
--- a/src/assets/sass/protocol/components/_notification-bar.scss
+++ b/src/assets/sass/protocol/components/_notification-bar.scss
@@ -5,6 +5,7 @@
 @import "../includes/lib";
 
 .mzp-c-notification-bar {
+    @include border-box;
     @include clearfix;
     @include text-body-sm;
     background-color: $color-marketing-gray-20;
@@ -13,13 +14,12 @@
     box-shadow: $box-shadow-md;
     color: $color-ink-80;
     font-weight: normal;
-    margin: $layout-xs auto 0;
-    text-align: center;
-    position: relative;
+    margin: $layout-xs $spacing-md 0;
     padding: $spacing-md $spacing-2xl;
+    position: relative;
+    text-align: center;
 
     @media #{$mq-sm} {
-        min-width: $content-sm;
         max-width: $content-md;
     }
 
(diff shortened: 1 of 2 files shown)


src/assets/sass/protocol/components/_notification-bar.scss
```diff
@@ -13,13 +14,12 @@
...
-    margin: $layout-xs auto 0;
-    text-align: center;
-    position: relative;
+    margin: $layout-xs $spacing-md 0;
     padding: $spacing-md $spacing-2xl;
+    position: relative;
+    text-align: center;
 
     @media #{$mq-sm} {
-        min-width: $content-sm;
```
stephaniehobson: 🔥

src/assets/sass/protocol/components/_notification-bar.scss
```diff
@@ -61,20 +62,23 @@
...
-        margin: 5px;
+        width: 20px;
 
-        &:hover,
-        &:focus {
+        &:hover {
             cursor: pointer;
         }
 
+        &:focus {
```
stephaniehobson: 🔥 

Can we add `:hover` everywhere there's a `:focus`?

PR #561, opened 2020-04-03
Implement sticky navigation (Fixes #560)

### LINKED ISSUE #560: Implement sticky navigation feature

We have a request to make the navigation organism support an optional feature for sticky scrolling. This change comes out of a CRO experiment that is captured in the epic here: https://github.com/mozilla/bedrock/issues/8676

Add an option to make the navigation component sticky, so that it remains visible on scroll.

### Please describe why these changes need to be made

The CRO experiment showed a significant increase in conversions as a result of sticky navigation, so there is a request to roll this out site wide on www.mozilla.org.

### Please provide any designs or prototypes of the proposed changes

Design discussions and prototypes happened here: https://github.com/mozilla/bedrock/issues/8674

It's the primary navigation on nearly every page of www.mozilla.org. The sticky behavior will be optional and disabled by default, so the risk to other sites should be minimal unless they opt-in to the new feature.

N/A

N/A

Sticky behavior will happen for desktop sized viewports only, and also requires a minimum viewport height.

Yes - we need to implement this in Protocol before bedrock can make use of it downstream.

alexgibson:
## Description

Implement sticky scrolling behaviour for navigation component as an opt-in feature

### Issue

#560

### Testing
Demo: https://demo2--mozilla-protocol.netlify.com/demos/navigation.html

- [x] Navigation should be sticky only when opt-in is specifically given.
- [x] Navigation should only be sticky when large viewport requirements are met.
- [x] Link clicks should still function as expected when nav is sticky.
- [x] Scrolling part way down from the top of the viewport should hide the sticky nav.
- [x] Scrolling back up the viewport should immediately show the sticky nav again.
- [x] Resizing the viewport between large/small screen sizes should not introduce bugs or break expected behaviour.
- [x] Sticky nav should still be accessible via keyboard when hidden (focus should bring it back into view).

### DIFF
--- a/src/assets/sass/protocol/components/_navigation.scss
+++ b/src/assets/sass/protocol/components/_navigation.scss
@@ -51,6 +51,34 @@
 }
 
 /* -------------------------------------------------------------------------- */
+// Sticky navigation styles
+
+@supports (position: sticky) {
+    html.mzp-has-sticky-navigation {
+        .mzp-c-navigation.mzp-is-sticky {
+            @include transition(transform 300ms ease-in-out);
+            left: 0;
+            position: sticky;
+            top: 0;
+            z-index: 1000;
+
+            &.mzp-is-scrolling {
+                box-shadow: 0 0 5px 1px rgba(9, 32, 77, .12);
+            }
+
+            &.mzp-is-hidden {
+                @include transform(translate(0, -100%));
+            }
+        }
+
+        // add scroll-offset for anchor links.
+        .mzp-is-anchor-link {
+            scroll-margin-top: 120px; /* stylelint-disable-line property-no-unknown */
+        }
+    }
+}
+
+/* -------------------------------------------------------------------------- */
 // Mozilla Logo
 
 .mzp-c-navigation-logo {
--- a/src/assets/sass/protocol/components/_navigation.scss
+++ b/src/assets/sass/protocol/components/_navigation.scss
@@ -63,11 +63,12 @@
             z-index: 1000;
 
             &.mzp-is-scrolling {
-                box-shadow: 0 0 5px 1px rgba(9, 32, 77, .12);
+                // Shadow colors are equivalent to $color-ink-90, $color-blue-90, $color-ink-90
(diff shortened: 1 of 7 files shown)


src/assets/sass/protocol/components/_navigation.scss
```diff
@@ -50,6 +50,34 @@
...
+    html.mzp-has-sticky-navigation {
+        .mzp-c-navigation.mzp-is-sticky {
+            @include transition(transform 300ms ease-in-out);
+            left: 0;
+            position: sticky;
+            top: 0;
+            z-index: 1000;
+
+            &.mzp-is-scrolling {
+                box-shadow: 0 0 5px 1px rgba(9, 32, 77, .12);
```
craigcook, line 66: We have tokens for box-shadows, though they're bigger than this and spread in all directions so I don't think a token can work here. But maybe we should at least match the color and layering of other shadows.

```suggestion
                // Shadow colors are equivalent to $color-ink-90, $color-blue-90, $color-ink-90
                box-shadow: 0 0 6px 1px rgba(29, 17, 51, .04), 0 0 8px 2px rgba(9, 32, 77, .12), 0 0 5px -3px rgba(29, 17, 51, .12);
```

src/assets/sass/protocol/components/_navigation.scss
```diff
@@ -50,6 +50,34 @@
...
+            }
+
+            &.mzp-is-hidden {
+                @include transform(translate(0, -100%));
```
craigcook, line 70: The shadow is still visible since it extends past the height of the nav, so we should make this a bit more than 100% to compensate. Maybe just `-101%` or we could even do `calc(-100% - 8px)`

alexgibson:
Thanks @craigcook updated. Yeah, we're not owrrying about sub navigation currently since we don't yet have a proper component for this in Protocol.

src/assets/sass/protocol/components/_navigation.scss
alexgibson: Seems like `-105%` seems to be the right amount. I went with a static value, as opposed to throwing `calc()` insto the mix.

PR #562, opened 2020-04-03
Updated button styles (#429)

### LINKED ISSUE #429: Add Firefox themed buttons
See: https://docs.google.com/document/d/1NgIKryJ0dCoPOGqE5A6rFgjDomd-PuVuLulFhPkN3N8/edit?usp=sharing

components-buttons

Buttons comes in 4 sizes | Small, Medium, Large and Extra Large.
-- | --
Buttons have 5 states | Default, Hover, Pressed, Focus and Disabled.
Buttons are accessible | Color Contrast, Typography, Interaction
There is a primary action button | This button is used as a primary call to action.
The primary button can be used on both light and dark backgrounds. | The primary button can be used on both light and dark backgrounds.
There is an inverted Secondary button. | This can be used on a light background.
There is an inverted Secondary Button that goes over a dark background. | This is used as an alternative for a button over a dark background.

stephaniehobson:
## Description

Updates button styles.

- [X] I have documented this change in the design system.

### Issue

Fix #429 

### Testing

http://localhost:3000/patterns/atoms/buttons.html
https://demo1--mozilla-protocol.netlify.com/patterns/atoms/buttons.html

### DIFF
--- a/src/assets/sass/protocol/components/_button.scss
+++ b/src/assets/sass/protocol/components/_button.scss
@@ -4,35 +4,100 @@
 
 @import '../includes/lib';
 
+$field-focus-ring: 0 0 0 2px rgba($color-blue-40, .5);
+$focus-border-color: #008aea;
+
+@mixin default-states {
+    &:focus {
+        border-color: $focus-border-color;
+    }
+
+    &:hover {
+        background-color: $color-marketing-gray-20;
+        border-color: $color-black;
+        box-shadow: none;
+        color: $color-black;
+    }
+
+    &:active {
+        background-color: $color-marketing-gray-20;
+        border-color: $color-marketing-gray-70;
+        color: $color-black;
+    }
+}
+
+@mixin default-states-dark {
+    &:focus {
+        border-color: $focus-border-color;
+    }
+
+    &:hover {
+        background-color: $color-dark-gray-60;
+        border-color: $color-white;
+        color: $color-white;
+    }
+
+    &:active {
+        background-color: $color-dark-gray-60;
+        border-color: $color-light-gray-50;
+        color: $color-white;
+    }
+}
+
+@mixin product-states {
+    &:focus {
+        border-color: $focus-border-color;
+    }
+
+    &:hover {
+        background-color: $color-blue-60;
+        border-color: transparent;
+        color: $color-white;
+    }
+
+    &:active {
+        background-color: $color-blue-80;
+        border-color: transparent;
+        color: $color-white;
+    }
+}
+
+@mixin product-states-dark {
+    @include product-states;
+}
+
 /* -------------------------------------------------------------------------- */
-// Primary Button
+// Button shape and size
 
 .mzp-c-button,
 a.mzp-c-button {
     @include border-box;
-    @include text-body-cta;
-    @include transition(background-color 100ms, box-shadow 100ms, color 100ms, transform 100ms);
-    background-color: $color-black;
-    border-radius: $border-radius-xs;
-    border: 2px solid $color-black;
-    color: $color-white;
+    @include font-size(16px);
+    border-radius: $border-radius-sm;
+    border: 2px solid transparent;
     cursor: pointer;
     display: inline-block;
     font-weight: bold;
-    padding: ($spacing-md - 2px) ($spacing-xl - 2px); //2px extra padding removed to compensate for 2px border.
+    padding: 6px (16px - 2px); //2px extra padding removed to compensate for 2px border.
     text-decoration: none !important;  /* stylelint-disable-line declaration-no-important */
 
-    &:focus,
-    &:hover {
-        @include transform(translate(-2px, -2px));
-        background-color: $color-white;
-        box-shadow: 4px 4px 0 0 $color-black;
-        color: $color-black;
+    &:focus {
+        box-shadow: $field-focus-ring;
+        border-color: $focus-border-color;
+        outline: none;
     }
 
+    &::-moz-focus-inner {
+        border: 0;
+    }
+
+    &:hover {
+        box-shadow: none;
+    }
+
+
     &:disabled {
-        background-color: $color-marketing-gray-30;
-        border-color: $color-marketing-gray-30;
+        opacity: 0.6;
         pointer-events: none;
     }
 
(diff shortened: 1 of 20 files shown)


src/assets/sass/protocol/components/_button.scss
```diff
@@ -4,35 +4,100 @@
 
 @import '../includes/lib';
 
+$field-focus-ring: 0 0 0 2px rgba($color-blue-40, .5);
```
craigcook, line 7: Should we use an opaque light blue instead of alpha? The transparency isn't very visible on a dark background but maybe that's fine since there's still the border. Or we could maybe make it a little more opaque just in the dark theme.

src/assets/sass/protocol/components/_button.scss
```diff
@@ -4,35 +4,100 @@
...
+$focus-border-color: #008aea;
```
craigcook, line 8: Should this be a token from the palette?

src/assets/sass/protocol/components/_button.scss
```diff
@@ -100,44 +187,77 @@ a.mzp-c-button.mzp-t-product {
...
+
+        &:active {
+            background-color: $color-marketing-gray-20;
+            border-color: $color-marketing-gray-50;
+            color: $color-marketing-gray-70;
+        }
+
+         // dark
+         &.mzp-t-dark {
+            background-color: rgba(255, 255, 255, 0.1);
```
craigcook, line 241: I think these could all use tokens like `rgba($color-white, .1)` (yay sass!)

src/assets/sass/protocol/includes/forms/_lib.scss
```diff
@@ -0,0 +1,81 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// Import all the things!
+@import '../lib';
+
+/* forms
```
craigcook, line 8: Should this file be in this PR or did it get copied over from the form work?

src/assets/sass/protocol/components/_button.scss
```diff
@@ -4,35 +4,100 @@
...
 .mzp-c-button,
 a.mzp-c-button {
     @include border-box;
-    @include text-body-cta;
-    @include transition(background-color 100ms, box-shadow 100ms, color 100ms, transform 100ms);
-    background-color: $color-black;
-    border-radius: $border-radius-xs;
-    border: 2px solid $color-black;
-    color: $color-white;
+    @include font-size(16px);
```
craigcook, line 75: Should this be `text-body-cta` or is that going away? I can't even remember...

src/assets/sass/protocol/components/_button.scss
```diff
@@ -4,35 +4,100 @@
...
+
 /* -------------------------------------------------------------------------- */
-// Primary Button
+// Button shape and size
 
 .mzp-c-button,
 a.mzp-c-button {
     @include border-box;
-    @include text-body-cta;
-    @include transition(background-color 100ms, box-shadow 100ms, color 100ms, transform 100ms);
```
craigcook, line 14: Should we keep the transition for the color changes?

src/assets/sass/protocol/components/_button.scss
craigcook: Oh I see, because of the new button sizes I don't think that single CTA style will cut it any more. So this is probably correct and maybe we do need to look into killing `text-body-cta`

src/assets/sass/protocol/components/_button.scss
```diff
@@ -41,53 +106,75 @@ a.mzp-c-button {
...
 
-.mzp-c-button.mzp-t-dark,
-a.mzp-c-button.mzp-t-dark {
-    background-color: $color-white;
-    border-color: $color-white;
-    color: $color-black;
+.mzp-c-button.mzp-t-small, // backwards compatability
+.mzp-c-button.mzp-t-md {
+    @include font-size(14px);
+    padding: 3.5px (16px - 2px);
```
craigcook, line 120: These seem like strange numbers. Can this just be `padding: $spacing-xs $spacing-md`?

stephaniehobson: The specs I have are for a specific button height. Which means that I can't jus- oh wait actually maybe I can. BRB.

src/assets/sass/protocol/includes/forms/_lib.scss
stephaniehobson: Yup, forgot to delete it.

stephaniehobson: Oh no, wait, I remember what I wanted it for.

src/assets/sass/protocol/components/_button.scss
stephaniehobson: I couldn't find a matching colour in the pallet.

src/assets/sass/protocol/components/_button.scss
stephaniehobson: I was following the specs Wendy provided but I agree it's hard to see. I'd be happy to just put it on a white matte (making it SUPER bright). But will defer to @justinkropp if he has an opinion

src/assets/sass/protocol/components/_button.scss
stephaniehobson: ✅

src/assets/sass/protocol/includes/forms/_lib.scss
stephaniehobson: Updated it to just include the relevant variables for the buttons and removed those variables from the button file.

src/assets/sass/protocol/components/_button.scss
stephaniehobson: Okay, I switched to `line-height` to define the height of these and I think that works better.

src/assets/sass/protocol/components/_button.scss
stephaniehobson: Yeah, it doesn't look like it is used by anything else. Do you want me to do that as part of this PR? That will make this a breaking change.

src/assets/sass/protocol/components/_button.scss
stephaniehobson: Can't hurt.

src/assets/sass/protocol/components/_button.scss
craigcook: Nah, we'll do that separately, maybe part of the theme work that includes updating the type scale already.

stephaniehobson: 👍 ✅

stephaniehobson:
Updated and rebased. But there's an outstanding question about the focus ring on dark backgrounds.

src/assets/sass/protocol/components/_button.scss
justinkropp: Good question. I think the border would suffice here, but perhaps the alpha could be bumped up to `.7` or `.8` and see if that helps.

stephaniehobson:
Screen Shot 2020-04-14 at 12 59 03

Top: current
Middle: less opacity
Bottom: same colour as on a white background

From an accessibility perspective the bottom one is the best.

@justinkropp thoughts?

justinkropp:
@stephaniehobson Let's move forward with the one on the bottom: _same color as on a white background._

stephaniehobson:
Update

src/assets/sass/protocol/components/_button.scss
craigcook: The greater line-height looks wonky if the button label wraps to two lines, which we can't guarantee will never happen. I don't think we need to exactly match the height in the spec to the exact pixel if we're at least close to it with whole numbers.

stephaniehobson: I think the greater line height looks fine.

craigcook: Medium button with the current `line-height: remify(32px - 4px);` (1.75rem) and `padding: 0 (24px - 2px);`:
image

With standard `line-height: 1.5` and `padding: $spacing-xs $spacing-lg;`
image

Maybe the buttons wrapping is an edge case. But it also seems like one we can solve for pretty easily and still avoid magic numbers in the CSS. If there's a good reason the medium button has to be *exactly* 32px tall and 33px is unacceptable then I suppose magic numbers are the only way. Other sizes might still need some padding math to compensate for the border width but if we're within a few pixels of the intended size I don't think anyone will complain.

justinkropp: @craigcook @stephaniehobson 

> If there's a good reason the medium button has to be exactly 32px tall and 33px is unacceptable then I suppose magic numbers are the only way.

We were using the pixel specific heights just so we can have something consistent to use amongst the designers and our files. No need to try to hack it or use magic numbers; if `33px` or something close, by all means use it. Hope that helps. :)

stephaniehobson:
Updated.

PR #593, opened 2020-05-27
Icons page

Errolyn:
## Description

Adds an Icon preview page for assets from the protocol-assets repo

### DIFF
--- a/gulpfile.js/tasks/drizzle.js
+++ b/gulpfile.js/tasks/drizzle.js
@@ -12,7 +12,7 @@ function renderIcons(iconData, depth=2){
         .forEach( (key) => {
             output += `<h${depth} class="icon-title">${key}</h${depth}>`;
             iconData[key].files.forEach( (file) => {
-                output += `<div class="img-group" data-clipboard-text="img/icons/${file.src}"><img src="/assets/protocol/protocol/img/icons/${file.src}" class="icons"/>`;
+                output += `<div class="img-group" title="Click to copy file path" data-clipboard-text="img/icons/${file.src}"><img src="/assets/protocol/protocol/img/icons/${file.src}" class="icons"/>`;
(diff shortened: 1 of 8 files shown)


craigcook:
This is awesome! I think we could probably spruce up the layout a bit but my nitpicks are fairly superficial. `overflow:scroll;` forces scrollbars even when they're not needed, so maybe `overflow:auto; would look better. I'm also not sure it needs the thick border on the label, and could use some spacing between boxes. Do you want to keep tinkering on the design solo or would a mockup help give you something specific to aim for?

src/assets/sass/docs/site.scss
```diff
@@ -345,3 +345,50 @@
...
+    width: 60px;
+    padding:3px;
+    border-radius: 6px;
+    background-color: $color-light-gray-30;
+}
+h2.icon-title {
+    @include font-base;
+    @include text-display-lg;
+}
+.img-group {
```
craigcook, line 362: Let's add `cursor: pointer;` to this rule to hint at clickability.

gulpfile.js/tasks/drizzle.js
```diff
@@ -3,9 +3,30 @@
...
+const Handlebars = require('handlebars');
+
+// Register handlebars helpers
+function renderIcons(iconData, depth=2){
+    let output = '';
+    Object.keys(iconData)
+        .forEach( (key) => {
+            output += `<h${depth} class="icon-title">${key}</h${depth}>`;
+            iconData[key].files.forEach( (file) => {
+                output += `<div class="img-group" data-clipboard-text="img/icons/${file.src}"><img src="/assets/protocol/protocol/img/icons/${file.src}" class="icons"/>`;
```
craigcook, line 15: We can add `title="Click to copy file path"` to offer a hint, along with the cursor change.

Errolyn:
Alright I added the suggested changes, and also a link to the github repo for the protocol-assets page because I realized that might be handy for people :D

PR #595, opened 2020-06-02
[fix #505] New content layout containers

### LINKED ISSUE #505: Content Layout: Adding a third size (S, MD, LG)
Right now we only have two sizes for content layout 736px wide and 1312px wide. This issue is there are times when we need a size in-between to make text areas more legible.

This text is too long which makes it harder to read/scan
Screen Shot 2019-10-09 at 9 43 46 AM

I am proposing we have 3 sizes S/M/L

- **S = 736** (this is for times we have a wall of copy "privacy notices")
- **MD = 992** (for those times illustrated in the image attached)
- **LG = 1248** (navigation, hero images and other areas where the text is large or minimal)

craigcook:
## Description
Adds container theme classes (deprecating `mzp-t-narrow`). Also finally actually documents the container in the first place.

### Issue
#505


src/pages/demos/content-container-vis.hbs
```diff
@@ -0,0 +1,26 @@
+---
+title: Visualization of content containers
+layout: blank
+styles:
+  - content-container
+---
+
+<div class="mzp-l-content mzp-t-content-sm display">
+  <p class="label"><code>.mzp-t-content-sm</code> <code>max-width: 464px</code></p>
```
stephaniehobson, line 9: 464 is the old value. I think this should be 432 now. Same with the other values.

craigcook: 🤦

craigcook:
Rebased and updated

src/assets/sass/protocol/base/elements/_sections.scss
```diff
@@ -4,18 +4,42 @@
...
-    @include border-box;
     @include clearfix;
     margin: 0 auto;
     max-width: $content-max;
     min-width: $content-xs;
     padding: $layout-xs;
     position: relative;
 
-    &.mzp-t-narrow {
+    &.mzp-t-content-sm {
```
stephaniehobson, line 17: My personal preference would be to put the default class styles before the modifier styles in the source order. (so move the two media queries on lines 45 and 49 up above this line).

Totally optional, personal, nit-pic and I realize you just used the existing source order when editing so definitely not a blocker.

craigcook: Well, the true default is without a modifier... but that then raises the question of whether we should stick to the standard of "medium" being the default and make a modifier for `.mzp-t-content-max`

craigcook: Oh I follow now, you mean the media queries at the end... yeah I should move those.

Issue #580, opened 2020-05-08
Add button alignment options


stephaniehobson:
Integrate the awesome work done by the SUMO team on a component for buttons:

<img width="1205" alt="button-wrap" src="https://user-images.githubusercontent.com/854701/81453220-cefde600-913d-11ea-955a-cda6b642a89d.png">

PR #601, opened 2020-06-12
Add button container and alignment options (#580)

craigcook:
## Description

Adds a button container element for forms, with a few alignment options. Also adds styling for the small button info text.

This is built on @stephaniehobson's work, which was in turn built on the work @kkellydesign did on SUMO.

### Issue

#580 

### Testing

https://demo1--mozilla-protocol.netlify.app/patterns/molecules/forms.html#button-container
https://demo1--mozilla-protocol.netlify.app/demos/forms/button-alignment.html

### DIFF
--- a/src/assets/sass/protocol/components/forms/_button-container.scss
+++ b/src/assets/sass/protocol/components/forms/_button-container.scss
@@ -33,7 +33,7 @@
         margin-top: $info-v-spacing;
     }
 
-    &.mzp-t-stretch .mzp-c-button {
+    &.mzp-l-stretch .mzp-c-button {
         max-width: none;
     }
 
@@ -57,7 +57,7 @@
             justify-content: flex-end;
         }
 
-        &.mzp-t-stretch {
+        &.mzp-l-stretch {
             align-items: stretch;
 
             .mzp-c-button {
--- a/src/assets/sass/protocol/components/forms/_button-container.scss
+++ b/src/assets/sass/protocol/components/forms/_button-container.scss
@@ -33,6 +33,10 @@
         margin-top: $info-v-spacing;
     }
 
+    &.mzp-t-stretch .mzp-c-button {
+        max-width: none;
+    }
+
     @media #{$mq-sm} {
         @include bidi(((text-align, left, right),));
         align-items: start;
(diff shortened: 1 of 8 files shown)


src/assets/sass/protocol/components/forms/_button-container.scss
```diff
@@ -0,0 +1,63 @@
...
+        max-width: $content-xs;
+        width: 100%;
+    }
+
+    // no bottom spacing if it's the last thing in the form
+    .mzp-c-form > .mzp-c-button-container:last-child {
+        margin-bottom: 0;
+    }
+
+    // align center
```
stephaniehobson, line 25: If center align is the default do we also need a class for it?

craigcook: It's centered by default on mobile, this makes it centered on all screens. Though if we don't even need that option then we could lose this class, too.

craigcook:
hmm... well buttons should only be full width if the form is less than X wide (not sure what X is, maybe 400px?). But container queries aren't a thing yet. So if we include a stretch class we have to at least add a strongly worded caution that it should never be used in any form that doesn't have a max-width of X and just trust people not to abuse it.

My other idea is to standardize a few widths for the form container, then we can hang the button width from the a class like `.mzp-c-form.mzp-t-width-sm .mzp-c-button { width: 100%; max-width: none; }` (or something like that) but that's additional work beyond the scope of this.

I don't trust people so I like guard rails, but we can't keep all our babies in plastic bubbles forever. I guess I'll add a stretch class.

craigcook:
Added stretch, but as a "t" class instead of "l" because it changes the style and behavior of the buttons, not just the layout... but I could be splitting hairs. Maybe they should all be "t"?

src/assets/sass/protocol/components/forms/_button-container.scss
stephaniehobson: I think we should keep the all screens option 👍

stephaniehobson:
I've always thought of "t" as how the button looks (colour, background, border, border radius) and "l" as how it takes up space on the page. By those criteria stretch is a "l". You seem to think about it differently though? What do you mean when you say it changes the style of the buttons?

PR #605, opened 2020-06-23
Publish 12.0.0

craigcook:
Bumps version to 12.0.0 (at last!)

PR #606, opened 2020-06-23
Icons page: update intro, de-list from nav

craigcook:
The new icons page is super-handy but when folks on the brand and UX teams saw it, it sparked a bigger conversation about updating our icon set and writing some usage guidelines. The consensus was that we shouldn't encourage people to use the icons right now because they're going to change soon.

So this change doesn't delete the page but removes it from the side nav until we have a chance to refine the icons and improve documentation (you can still reach it directly if you know the URL) and adds a stronger warning.

### DIFF
--- a/src/pages/fundamentals/icons.hbs
+++ b/src/pages/fundamentals/icons.hbs
@@ -1,16 +1,13 @@
 ---
-title: Icon Assets
-name: Icon Assets
-order: 5
+name: Icons
 labels:
   - inprogress
 ---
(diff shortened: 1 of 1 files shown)


craigcook:
Blocks https://github.com/mozilla/protocol/pull/605

src/pages/fundamentals/icons.hbs
```diff
@@ -1,16 +1,13 @@
 ---
-title: Icon Assets
```
craigcook, line 2: Note for reviewer: Omitting the title effectively hides this in the nav (the link is still there, it's just empty). This is a total hack but I hope it's short-lived.

Issue #410, opened 2019-07-15
Update type scale(s)


wboucherfischer:
On https://protocol.mozilla.org/fundamentals/typography.html please update to show Custom Headings that are taken from the Brand Guidelines, in addition to showing default. 

**All are in Metropolis Bold**

- Display XXL - Font Size, 64px (4rem), Line Height 72px
- Display XL - 56px (3.5 rem), Line height 64px
- Display LG - Font Size - 48px (3rem), Line Height, 56px
- Display MD - 40px (2.5rem), Line Height - 44px
- Display SM - 32px (2rem), Line height - 36px
- Display XS - 24px (1.5rem), Line height: 28px

See: https://docs.google.com/document/d/1NgIKryJ0dCoPOGqE5A6rFgjDomd-PuVuLulFhPkN3N8/edit?usp=sharing

craigcook:
All of these sizes are already in Protocol, or is this just about updating the line heights? Current Protocol uses a 1.1 line-height for all headings whereas the brand guide rounds them to the nearest multiple of four. Protocol used to do that same rounding before we simplified it, but we can bring it back.

wboucherfischer:
We have the default headings (h1, h2, h3, h4, h5, h6), but we had talked about custom heading classes to call. I think those should be defined and on this page. So if somebody wants to make an h1 smaller, etc. I think a 1.1 line height is fine.

craigcook:
PR #387 relates to this (which in turn references issue #297)

But I think we might need to discuss the line height thing further and try to reach consensus, and probably separately from this issue. I'll file another issue for that.

craigcook:
Rambling thoughts incoming.

So... most of what we've been talking about in relation to atomic headings is only regarding the sizing, which we already do with mixins. We can make any element any size with a line of CSS, with no further effort required.

We can instead create a set of presentational utility classes (as in #387 ), so we could arbitrarily size elements in HTML without writing custom CSS for it, but that upsets my purist sensibilities so I'd like to avoid that. Using classes for sizing means we use HTML to describe appearance. We could make pages without ever writing any page-level CSS, but I'm not sure that's actually a good goal. Sizing a heading with a utlity class instead of a mixin isn't really turning it into an "atomic component", it's just a shortcut to avoid writing new CSS. If we have to write HTML to specify a size, I'm not sure `class="mzp-u-display-xl;"` is any better than `style="font-size: 2rem;"`.

[...]

Issue #405, opened 2019-07-10
Update Firefox theme text colors


craigcook:
Once the color tokens have been updated in https://github.com/mozilla/protocol-tokens/pull/37, the Firefox theme should use `#42425a` as the base text color and `#20133a` for headings.

wboucherfischer:
Please update to include this on https://protocol.mozilla.org/fundamentals/color.html to show as 'Display Font Primary Colors'.

See: https://docs.google.com/document/d/1NgIKryJ0dCoPOGqE5A6rFgjDomd-PuVuLulFhPkN3N8/edit?usp=sharing

johngruen:
#447

Issue #515, opened 2019-10-28
Update defaults for mzp-t-dark


stephaniehobson:
In our links file we make links in `mzp-t-dark` components white. But we don't change the default text colour or the background colour anywhere. 

This means that the theme class cannot be used to alter the theme of components by itself and can only be used with components which support it.

In bedrock when we use the dark theme we have to declare styles like:
```
.mzp-t-dark {
    background-color: $color-off-black;
    color: $color-white;
}
```

And in our IE fallback sheet we have:
```
// make sure dark components have correct background and text color
.mzp-t-dark {
    background: $color-off-black;
    color: #fff;

    h1, h2, h3, h4, h5, h6 {
        color: #fff;
    }
}
```

This theme should produce readable text when it is applied. Styles can then be tweaked from there.

I suggest:
```
.mzp-t-dark {
    background-color: #000000;
    color: $color-white;

    .mzp-t-firefox & {
          background-color: $color-off-black;
    }
}
```

I think this will be a breaking change and I will do some testing in bedrock to see if this is possible at all.

stephaniehobson:
A survey of bedrock suggests this will break more than a few pages:

- firefox/privacy/products
- welcome
- developer/firstrun
- firefox/concerts
- firefox

It feels like the right thing to do to set a background colour when the foreground colour is changed. Is it worth the pain?

stephaniehobson:
Another option would be to not define the link colour in the links file and require it to be declared specifically on supporting components.

That will break things in different ways. Ways that would be much faster/simpler to fix than applying a background colour by default.

stephaniehobson:
A Slack channel conversation on this turned up a vote for removing the default link change.

craigcook:
https://github.com/mozilla/protocol/issues/853

PR #609, opened 2020-06-30
[fix #447] Implement brand themes

### LINKED ISSUE #447: (9/26) Implement theming
In order to manage application of Protocol as it applies to Mozilla and Firefox design systems we will need to create a set of theming conventions that cascade down from a top level declarations.

## Requirements
* typographic sizing, color, and font classes should be separate. This separation will allow for down-the-road flexibility for creating different colorways, type scales, and font-usage patterns.
* in order to close this issue we should implement the following:

1. a class that implements the new firefox font stack and default document color scheme w/in a document
2. a class that implements the mozilla font stack and default document color scheme w/in a document
3. a class that implements the firefox type scale w/in a document
4. a class that implements the mozilla type scale w/in a document

## Recommendations

right now, there's an `mzp-t-firefox` class used to set Open Sans...i think we need to be a little more granular about the naming system here, something like this might work: 
```
<body class="mzp-t-brand-firefox mzp-t-usage-presentation">
```

In this version any color and font stack selectors of can be descendants of the `mzp-t-brand-firefox` class, and all scaling selectors can be descendants of of the `mzp-t-usage-presentation` class.

The reason for separating these is because once Protocol is used for app UI, it is very likely that the overall scale will change, so i could easily see a variant here with `mzp-t-usage-application` that keeps the Firefox theme but wants different sizings (and, for example, no `:visited` color on links).

Additionally, this leaves room for another optional base class (`mzp-t-colorway-inverted` or similar) that could override a color theme.

craigcook:
## Description

This finally adds basic theming to Protocol. Themable styles are assigned to variables defined in maps, then the `get-theme()` function can swap to a different map based on the value of a single theme variable.

This also adds an alternative condensed type scale that works somewhat similarly, swapping to a different map of sizes based on another variable. 

### Issue
#447 - Implement theming
#405 - Update Firefox theme text colors
#410 - Update type scales

### Testing
https://demo1--mozilla-protocol.netlify.app/

This is a pretty big PR that touches a lot of files, but a lot of that is just tidying up comments and tweaking some docs. The real meat of the theming is in `includes/_themes.scss` and `includes/_functions.scss`


src/assets/sass/protocol/_theme-config.scss
```diff
@@ -0,0 +1,45 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// This file defines global theme settings for your project. You can
```
craigcook, line 5: Note for reviewer: I put this file in the root because I want to discourage consumers from editing any other files, especially since they're subject to change on future updates. Otherwise we could have defined these variables at the top of _lib, but I really don't like the idea of people messing with that file.

And defining them in protocol.scss wouldn't really work because it's really _lib that needs to know these values (they have to be defined _before_ _lib is imported) but maybe there's a better way.

stephaniehobson:
We should publish a point release to fix this: https://github.com/mozilla/protocol/issues/611 before merging the theme changes. I'll still try to start the review though.

src/assets/sass/protocol/_theme-config.scss
stephaniehobson: I think this is the best way to do it 👍

src/assets/sass/protocol/includes/_functions.scss
```diff
@@ -23,3 +27,95 @@
...
+}
+
+
+// Get a theme variable from the collected map.
+// This swaps the map based on the value of $brand-theme.
+// Theme maps are defined in includes/_theme.scss
+
+@function get-theme($var) {
+  @if not $var {
+    @return false;
```
stephaniehobson, line 100: Error handling here too?

src/assets/sass/protocol/base/elements/_document.scss
```diff
@@ -7,18 +7,23 @@
...
-    background: #fff;
+    background: $color-white;
 }
 
 body {
     @include text-body-md;
-    background: #fff;
-    color: $color-black;
-    font-family: $font-base-family;
+    background: $color-white;
```
stephaniehobson, line 15: get-theme?

src/assets/sass/protocol/_theme-config.scss
stephaniehobson: Wait, can this file be ridden with a local copy? Or are they going to have to edit a file in the npm package?

craigcook: Crap you're right... it should be overridden locally, which means I need to rethink this... maybe we don't need a config file at all and we can just set the defaults at the top of _lib after all. I'll work on it.

stephaniehobson: Well, these are identified as "default" in the config so the values should be over-rideable by a later declaration. In bedrock could we do it in the same place we do these definitions?

```
$font-path: '/media/fonts';
$image-path: '/media/protocol/img';
```

craigcook: Yeah, that's how it works, but I wrote out all these comments as if someone would modify their local copy... which you could totally do if you're just copying Protocol (like I was doing for all my testing), but when it's used as a package and this file lives in a node_modules folder nobody should be editing it and these comments are a bit misleading.

We do still have to declare _some_ default, but we can just put it right in _lib instead of in this separate file since we don't expect people to edit it, and I'll work on updating the comments and theme docs to explain how/where to change themes in a local file.

src/assets/sass/protocol/includes/_functions.scss
```diff
@@ -23,3 +27,95 @@
...
+// to convert to rem with px fallback. Example:
+//    @include font-size(type-scale('title-xl-size'));
+//
+// An alternative, condensed type scale is defined in a separate
+// map. When the $type-scale variable is set to 'condensed' this
+// function will draw its values from that alternative map.
+
+@function type-scale($scale) {
+  @if not $scale {
+    @return false;
```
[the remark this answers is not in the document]
craigcook: Looks like Sass already throws an error if the argument is missing so this whole `@if not $scale` bit isn't really needed, we get error handling for free.

src/assets/sass/protocol/base/elements/_forms.scss
```diff
@@ -109,27 +93,24 @@ fieldset {
 
 legend {
     @include text-body-lg;
+    font-family: get-theme('title-font-family');
     font-weight: bold;
     margin-bottom: $field-v-spacing;
-
-    .mzp-t-firefox & {
```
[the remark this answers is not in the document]
craigcook: I didn't find it in any of the form markup but there were plenty of now-redundant instances elsewhere that I've removed.

src/assets/sass/protocol/base/_elements.scss
```diff
@@ -11,4 +15,17 @@
...
 @import 'elements/tables';
-@import 'elements/typography';
+@import 'elements/titles';
+
+
+p,
+ul,
+ol,
+dl,
+table {
```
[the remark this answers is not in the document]
craigcook: Moved to a new _common.scss file

craigcook:
Updated.

src/assets/sass/protocol/components/forms/_form.scss
```diff
@@ -56,36 +56,27 @@ Basic atoms that make up a form
     }
 }
 
-// heading
 
+// Heading/title
```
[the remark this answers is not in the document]
craigcook: Now that you mention it, do we want to rename this to `.mzp-c-form-title`?

craigcook:
Temporarily putting this on hold because we have a few other small things in the pipe that we want to get into bedrock quickly, so it'll be better to merge them ahead of this PR so they don't get tied to a major release.

src/assets/sass/protocol/base/elements/_document.scss
```diff
@@ -7,18 +7,23 @@
...
-    font-family: $font-base-family;
+    background: get-theme('background-color');
+    color: get-theme('body-text-color');
+    font-family: get-theme('body-font-family');
+    line-height: type-scale('body-line-height');
     -moz-osx-font-smoothing: grayscale;
     -webkit-font-smoothing: antialiased;
 }
 
+.mzp-t-dark {
```
stephaniehobson, line 23: Declaring a foreground colour without a background colour has always made me uncomfortable. Last time we talked about it we agreed to remove it: https://github.com/mozilla/protocol/issues/515

If you're going to add the text colour change here it needs a background color declaration as well, and then we need to deal with the fallout from that.

I think this is a good time to do that actually since this is a breaking change.

If you'd like I can put together a separate PR with the mzp-t-dark changes to go live with the theme changes.

src/assets/sass/protocol/components/forms/_form.scss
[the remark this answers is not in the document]
stephaniehobson: Yes.

src/assets/sass/protocol/base/elements/_document.scss
craigcook: I'm removing this. We don't really have a proper global dark theme, it's just been done on a per-component basis so I think adding such a sweeping rule here does more harm than good. And once I remove this I'll double-check all the components with dark versions.

craigcook:
Believe it or not, this is ready for review again (!).

craigcook:
Once again marking Do Not Merge temporarily so we can get a small update published first (#638)

src/assets/sass/protocol/includes/_lib.scss
```diff
@@ -2,8 +2,49 @@
...
+//     'standard'
+//     'condensed'
+
+$type-scale: 'standard' !default;
+
+
 // Import all the things!
 @import './node_modules/@mozilla-protocol/tokens/dist/index.scss';
-@import 'variables';
+@import 'themes';
```
stephaniehobson, line 48: Do we not also need variables?

stephaniehobson (approved):
R+ 🎭 

I suggest we make it Mozilla out of the box instead of Firefox out of the box. It's not a blocker if you disagree but it's a decision we have to live with going forward so now's the time to consider it.

craigcook:
Updated!

PR #612, opened 2020-07-08
[fix #611] Reduce content container min-width

### LINKED ISSUE #611: mzp-l-content too wide on small screens
As a result of the new content sizes which use `box-sizing:content-box` content now causes crawlbars on small screens.

craigcook:
## Description

The border-box declaration was removed in #595 but it probably shouldn't have been. This puts it back.

### Issue

#611 

### Testing
Test responsiveness in small viewports. There should be no crawlbars as far down as 320px (smaller than that still crawls but that's OK).

http://localhost:3000/demos/content-containers.html
http://localhost:3000/demos/content-container-vis.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# HEAD
+
+## Bug Fixes
+
+* **css:** Reduce content container min-width to avoid crawlbars (#611)
+
 # 12.0.0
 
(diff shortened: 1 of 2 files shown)


craigcook:
Re-removed the border-box and instead made the min-width smaller.

PR #621, opened 2020-07-14
Remove max-width from cards in card-layout (Fix #620)

### LINKED ISSUE #620: Card half doesn't fill half width of large screens
86477438-5895f280-bd40-11ea-85b3-4eb05259789a

stephaniehobson:
## Description

Remove max-width from cards in card-layout.

### Issue

Fix #620

### Testing

- This is a very brittle way of doing this, it depends on the card template CSS to be included after the card CSS. Should I increase the specificity to avoid this problem?
- I am wondering if we should divorce the card widths from the card sizes all together. Perhaps that is a bigger issue that should be considered separately. Do we ever use the cards without the layout options?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Bug Fixes
 
+* **css:** Remove max-width from cards in card-layout (#620)
 * **css:** Navigation is max-content width on large screens (#615)
 * **css:** Set text-align on image replacement mixin (#618)
 * **css:** Reduce content container min-width to avoid crawlbars (#611)
--- a/src/assets/sass/protocol/components/_card.scss
+++ b/src/assets/sass/protocol/components/_card.scss
@@ -126,7 +126,7 @@
     }
 
     @media #{$mq-sm} {
-        max-width: 450px;
+        max-width: $content-sm;
     }
 
     @media #{$mq-md} {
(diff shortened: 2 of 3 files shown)


craigcook:
> * This is a very brittle way of doing this, it depends on the card template CSS to be included after the card CSS. Should I increase the specificity to avoid this problem?

I'm not sure this is true. I just tried in a demo and it seems to work the same no matter what order I import them. But I might be missing something, I need to test more thoroughly.

> * I am wondering if we should divorce the card widths from the card sizes all together. Perhaps that is a bigger issue that should be considered separately. Do we ever use the cards without the layout options?

I've had similar thoughts. I've never liked having cards described by their size when they're otherwise the identical component. It would be nice to just have "a card" and its size is determined by its container, but I'm not sure how practical that really is. I know part of the reason we ended up with specific sizes was due to keeping the image aspect ratios, but maybe we can figure out a better approach... but that would indeed be bigger than this bug fix.

craigcook:
> I'm not sure this is true. I just tried in a demo and it seems to work the same no matter what order I import them. But I might be missing something, I need to test more thoroughly.

I was wrong, they do need to be imported in order. 😭

stephaniehobson:
Sprinkled on a bit more specificity.

Issue #430, opened 2019-07-17
Add Firefox themed text input [forms]


wboucherfischer:
See: https://docs.google.com/document/d/1NgIKryJ0dCoPOGqE5A6rFgjDomd-PuVuLulFhPkN3N8/edit?usp=sharing

Normal Text Input | Default text input
-- | --
Text Input, Hover | When user hovers over text box, a hover displays.
Text Input, Focus | When the user has selected the text box, a focus selection appears.
Text Input, Error | When there is an error in the text box element, an error state shows.
Text Input, Success | When the user is successful in enter an item in a text box, a success state shows.
Text Input, Disabled | There is a disabled state for the text box.

![text-input](https://user-images.githubusercontent.com/50332473/61409660-8e279e80-a897-11e9-911f-1892f75f5310.png)

dzingeek:
Here are the visual specs
![input-field-specs](https://user-images.githubusercontent.com/3740746/67119316-c0e57980-f19b-11e9-91c0-6f941f8c3e6d.png)

dzingeek:
@stephaniehobson here are the icons for the input fields (check, alert, down arrow...)
[input-field-icons.zip](https://github.com/mozilla/protocol/files/3795333/input-field-icons.zip)

dzingeek:
**Here is an update to the error state**
![input field specs](https://user-images.githubusercontent.com/3740746/69278357-016b5500-0b97-11ea-86f1-b6aafbc67b49.png)

dzingeek:
Just realized I didn't spec out the box shadow

[**.elevation-02** {](https://github.com/mozilla/protocol/issues/456)
  box-shadow: 0 16px 24px  2px hsla(261, 50%, 13%, .04),
              0 6px  32px  4px hsla(220, 79%, 17%, .12),
              0 8px  12px -5px hsla(261, 50%, 13%, .12);

craigcook:
> Just realized I didn't spec out the box shadow
> 
> [**.elevation-02** {](https://github.com/mozilla/protocol/issues/456)
> box-shadow: 0 16px 24px 2px hsla(261, 50%, 13%, .04),
> 0 6px 32px 4px hsla(220, 79%, 17%, .12),
> 0 8px 12px -5px hsla(261, 50%, 13%, .12);

That shadow looks like this:
![image](https://user-images.githubusercontent.com/205591/69284487-970ce180-0ba3-11ea-8f80-7f41e5604bc5.png)

[...]

PR #631, opened 2020-08-07
Add error message component

stephaniehobson:
## Description

Add error message component and update documentation associated with form validation.

### Issue

Error messages are defined in https://github.com/mozilla/protocol/issues/430#issuecomment-556399694

### Testing

- http://localhost:3000/demos/forms/validation.html
- http://localhost:3000/patterns/molecules/forms.html#message
- http://localhost:3000/patterns/atoms/forms.html#error-list


craigcook (commented):
This all looks great, my only nitpick is that the red background might be too much. Maybe it's just my aging eyes but I find it hard to read the small white text on the bright red:

image

We could go a little darker for better contrast maybe?

stephaniehobson:
Updated.

src/assets/sass/protocol/includes/forms/_lib.scss
```diff
@@ -80,3 +81,29 @@ $line-height-shim: 0.15em; // two elements with text appear to have more space b
...
+    margin: 2px 0 0;
+    max-width: 100%;
+    padding: $spacing-xs #{$spacing-sm + 2px};
+    position: relative;
+}
+
+@mixin form-msg-pointer() {
+    @include bidi(((left, 6px, right, auto),));
+    border: 7px solid transparent;
+    border-bottom: 8px solid $form-red;
```
craigcook, line 102: The pointer should be the same color as the background.

craigcook (approved):
The darker red is definitely better contrast, just need to update the little pointer thingie to match. r+wc! 🥠

stephaniehobson:
Updated.

PR #639, opened 2020-10-20
Fix #638 - Revise the sticky promo to be hidden by default 

### LINKED ISSUE #638: Add open functionality to Sticky Promo component

In order to remember if this promo has been dismissed, we need to set a cookie and then hide the promo by default. Rather than bake all that special functionality into Protocol, we want to make this require an open function (instead of just visible-by-default) 

Revise CSS so that this is hidden by default
Add open function to JS to show the sticky promo

/firefox/new pages

In order to assess the risk of changes, we need to know where components exist in the wild.

* When set in front of a dark background: N/A
* When set in front of a light background:  N/A
* When viewed in a mobile/responsive viewport: Same functionality as desktop 

Yes – https://github.com/mozilla/bedrock/issues/9406. If a user dismisses the sticky promo and refreshes/revisits the page, it shows up again.

maxxcrawford:
## Description

The sticky promo now requires an appended class to show. 
I also added a helper class to show the sticky promo on page load, providing backwards compatibility for previous behavior.

### Issue

https://github.com/mozilla/bedrock/issues/9406

### Testing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
-# 12.1.0
+# HEAD
+
+## Features
+
+* **css:** Revise sticky promo component to be hidden by default  (#638)
+
+# 12.0.1
 
 ## Features
 
--- a/src/assets/sass/protocol/components/_sticky-promo.scss
+++ b/src/assets/sass/protocol/components/_sticky-promo.scss
@@ -25,12 +25,7 @@ $logos: (
 }
 
 .mzp-c-sticky-promo {
-    @include bidi(((animation-name, mzp-a-slide-in-right, mzp-a-slide-in-left),));
-    @include bidi((
-        (right, $spacing-md, auto),
-        (left, auto, $spacing-md),
-    ));
-    animation: 600ms ease 1500ms both;
+
     background-color: $color-white;
     border-radius: $border-radius-sm;
     bottom: 0;
@@ -41,6 +36,15 @@ $logos: (
     position: fixed;
     width: $content-xs;
     z-index: 10;
+    animation: 600ms ease 1500ms both;
+
+    &.mzp-a-slide-in {
+        @include bidi(((animation-name, mzp-a-slide-in-right, mzp-a-slide-in-left),));
+        @include bidi((
+            (right, $spacing-md, auto),
+            (left, auto, $spacing-md),
+        ));
+    }
 
     &.mzp-a-fade-out {
         @include animation(mzp-a-fade-out 100ms ease 5ms both);
@@ -106,4 +110,3 @@ $logos: (
         background: transparent url("#{$image-path}/icons/close-white.svg") center center no-repeat;
     }
 }
-
--- a/src/assets/sass/protocol/components/_sticky-promo.scss
+++ b/src/assets/sass/protocol/components/_sticky-promo.scss
@@ -39,8 +39,12 @@ $logos: (
     position: fixed;
     width: $content-xs;
     z-index: 10;
-    animation: 600ms ease 1500ms both;
+    animation: 600ms ease 300ms both;
     opacity: 0;
+
+    &.mzp-js-show-on-load {
+        animation-delay: 1000ms;
+    }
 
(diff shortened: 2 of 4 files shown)


src/patterns/molecules/sticky-promo/sticky-promo.hbs
```diff
@@ -2,6 +2,8 @@
 name: Sticky Promo Card
 description: The primary purpose of the Sticky Card component is to ensure that the content it’s promoting remains a constant on the page -- always visible within the viewport no matter a user’s scroll position. Sticky Promo Cards can be used to promote a product and provide a CTA. While the card itself is both content and context agnostic, it should only be used strategically on pages where it will be most effective.
 notes: |
+  - The element is hidden by default. To show it, add `mzp-a-slide-in` to the class list.
```
craigcook, line 5: This instruction is confusing, and makes it sound like I should add that class to the HTML by hand. Really the way to display the promo is by firing `Mzp.StickyPromo.open()` with JS, and that function in turn adds the animation class.

So let's change this to something like "The element is hidden by default and made visible by the `Mzp.StickyPromo.open()` function."

src/patterns/molecules/sticky-promo/sticky-promo.hbs
```diff
@@ -20,6 +22,11 @@ notes: |
...
 </aside>
 
-<script
- src="../../../assets/protocol/protocol/js/protocol-sticky-promo.js">
+<script src="{{@root.baseurl}}/assets/protocol/protocol/js/protocol-sticky-promo.js"></script>
+
+<script>
+(function() {
+    'use strict';
+    Mzp.StickyPromo.open();
```
craigcook, line 30: When I make a button to show the promo onclick, then close it, then click to show it again, it remains invisible because it still has the `mzp-a-fade-out` class. I think the close event should reset it back to its original state.

src/assets/sass/protocol/components/_sticky-promo.scss
```diff
@@ -41,6 +39,12 @@ $logos: (
     position: fixed;
     width: $content-xs;
     z-index: 10;
+    animation: 600ms ease 1500ms both;
```
craigcook, line 42: This delay should only be when it's set to appear automatically. If I show it onclick (or any other event I might use to call the open function) I would expect it to appear right away and having the delay makes it feel sluggish.

I think we can do something like this:
```
&.mzp-js-show-on-load {
    animation-delay: 600ms;
}
```

craigcook: 1500ms also feels a little slow... it's a full second and a half, which is an eternity in UI response time. Though it feels ok being slower onload but it's sluggish onclick. So maybe we default to something like 500ms or 750ms, but when it's show-on-load we slow that down to 1500ms or even 1000ms.

craigcook (changes requested):

src/assets/sass/protocol/components/_sticky-promo.scss
maxxcrawford: Revised default `animation-delay` to `300ms` and the show-on-load to `1000ms`.

maxxcrawford:
@craigcook I removed the `Mzp.StickyPromo.open()` from the demo and added a button to click/attach the event. Thanks for the review!

craigcook (approved):
Looks good. Just add a note to the changelog and we can get this merged, then we'll do a version bump and publish.

CHANGELOG.md
```diff
@@ -1,4 +1,10 @@
-# 12.1.0
+# 12.1.1
```
craigcook, line 1: This should be `#HEAD` for this commit just to keep the history clean, and we'll version it in the other PR for publishing.

craigcook (approved):
Nice work! r+wc 🍍

CHANGELOG.md
maxxcrawford: Updating.

PR #649, opened 2020-12-11
Add white-links mixin [fix #648]

### LINKED ISSUE #648: Add a mixin for white links in components with dark backgrounds
We have a `light-links` mixin that applies inverted link colors for dark backgrounds. Previously it set links to white (for all states) but in 13.0.0 that changed to implement new link colors (including different colors for different states; light blue for `:link`, light purple for `:hover` etc). But we regularly want to set inverse links to white explicitly and the colored `light-links` mixin isn't always desirable.

We should add an alternative `white-links` mixin for this.

craigcook:
## Description
Adds a new mixin for explicitly white links (as opposed to the inverted theme colors).

### Issue
#648

### DIFF
--- a/src/assets/sass/protocol/components/_footer.scss
+++ b/src/assets/sass/protocol/components/_footer.scss
@@ -9,6 +9,7 @@
 
 .mzp-c-footer {
     @include font-base;
+    @include white-links;
     background: $color-black;
     clear: both;
     color: $color-white;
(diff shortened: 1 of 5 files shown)


src/assets/sass/protocol/components/_footer.scss
```diff
@@ -18,16 +19,14 @@
...
 
-    a:hover,
-    a:focus,
-    a:active {
-        color: $color-white;
-        text-decoration: none;
+        &:hover,
+        &:focus,
+        &:active {
+            text-decoration: underline;
```
alexgibson, line 28: Should hover / active link styles have some form of differential treatment here, like `text-decoration: none`? Right now the links in `mzp-c-footer-secondary` don't seem to have any real hover state.

alexgibson (changes requested):
Looks good, just one question on the default hover states for footer links?

src/assets/sass/protocol/components/_footer.scss
craigcook: Oops, this should be `text-decoration: none;`. Bad copypasta.

alexgibson (approved):

Issue #345, opened 2019-03-21
Finalize Updated Grid Sizing Rules


johngruen:
From the board:

> Protocol needs either a standardized grid layout system (at most), or a predefined set number of container sizes (at least).

Also include advice on two-column layout. Per Craig Cook: 

> We’ve seen some designs placing content in two columns, which isn’t something we have a ready-made technique for. There are specific components that arrange side by side content, and we have some templates to arrange components in columns, but we don’t have a general-purpose, content-agnostic way to put things in columns. This is a common design pattern for which we should have a reusable template.

Need to define text line length for 1-column layout.

johngruen:
Blocker for #179

mheubusch:
Current designs show 80 pixel units but there is not a corresponding token for this gutter.

justinkropp:
The proposed content-widths are as follows:

-     S = **432**
-     MD = **688**
-     LG = **928**
-     XL = **1152**

The number of columns can range anywhere from 1-column to 3-columns within these content widths. The space between columns for all sizes is 80px, except for the 688 content width, which gets a 64px width between columns.

**For reference:**

- ![768 Breakpoint + Content Width](https://user-images.githubusercontent.com/1461798/74279947-d3a33180-4ce9-11ea-9264-59aac79272c8.png)
- ![1024 Breakpoint + Content Width](https://user-images.githubusercontent.com/1461798/74279948-d43bc800-4ce9-11ea-82c7-fda072f8b0a3.png)
- ![1312 Breakpoint + Content Width](https://user-images.githubusercontent.com/1461798/74279949-d43bc800-4ce9-11ea-8a86-3b00fc2c6f27.png)

justinkropp:
@stephaniehobson I've got some updated visuals/assets that will help solidify our grid framework. I'll drop them here and we can talk about them.

stephaniehobson:
After discussing it with @craigcook we're going to use 64px for the 1024 breakpoint because otherwise the 4 column option for cards and pictos gets too squished and it'd be a lot of work to fix or remove those options.

PR #667, opened 2021-03-16
Add horizontal grid tokens #345

stephaniehobson:
## Description

- Adds variables to use when defining the horizontal grid for components. 
- Decreased padding on content containers at 1024px breakpoint (this is in preparation for updating components to use this larger grid).
- Updates some hard-coded values to use existing variables

---

### Issue

See #345

### Testing

Not much to see yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # HEAD
 
+## Features
+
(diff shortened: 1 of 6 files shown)


stephaniehobson:
Actually, I should add the vertical tokens since we have those too. And those should go in theme variables...

PR #669, opened 2021-03-17
Add horizontal and vertical spacing tokens #345 #536

### LINKED ISSUE #345: Finalize Updated Grid Sizing Rules

### LINKED ISSUE #536: Vertical Rhythm
Please use this template to request a new component.
If you're not sure whether a component already exists,
go ahead and file the issue here, and the team will triage appropriately.

Fill in as many sections as possible below.

What does the component do? In what contexts is it used. Be as specific as possible.

### Please provide any designs or previous implementations of the proposed component

Add screenshots, mockups, code prototypes, etc. in this section.

## Please describe what this component does in the following conditions

### Please provide examples of this type of component in use in non-Mozilla web properties

Add screenshots and links to web properties that use a similar component.
(We ask for this because we want to make sure we are following industry conventions while building components)

Generally speaking component definition should happen after a singleton instance has been developed,
or if the Protocol team has come to a consensus on need.

stephaniehobson:
## Description

- Adds a spacing them with horizontal and vertical spacing. 
- Using the new spacing decreases outside padding on content containers at 1024px breakpoint (this is in preparation for updating components to use the wider spacing).
- Updates some hard-coded values to use existing variables

---

### Issue

Fix #345
Fix #536

### Testing

Not much to see yet. You can compare http://localhost:3000/demos/content-containers.html to prod if you want.

### DIFF
--- a/src/assets/sass/protocol/includes/_themes.scss
+++ b/src/assets/sass/protocol/includes/_themes.scss
@@ -6,6 +6,34 @@
 // themes. Mixins and functions can draw from these variable maps, swapping to
 // a different map based on the value of the global $brand-theme variable.
 
+//* -------------------------------------------------------------------------- */
+// Vertical and Horizontal Spacing
+
+
+$theme-spacing-firefox: (
+    // xs
+    v-grid-xs: $layout-md,
+    h-grid-xs: $layout-xs,
+
+    // sm is the same as xs
+    v-grid-sm: $layout-md,
+    h-grid-sm: $layout-xs,
+
+    // md
+    v-grid-md: $layout-md,
+    h-grid-md: $layout-md,
+
+    // lg is the same as md
+    v-grid-lg: $layout-md,
+    h-grid-lg: $layout-md,
+
+    // xl
+    v-grid-xl: $layout-xl,
+    h-grid-xl: 80px,
+);
+
+$theme-spacing-mozilla: $theme-spacing-firefox;
+
 
 //* -------------------------------------------------------------------------- */
 // Fonts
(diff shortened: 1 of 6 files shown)


src/assets/sass/protocol/includes/_themes.scss
```diff
@@ -6,6 +6,34 @@
 // themes. Mixins and functions can draw from these variable maps, swapping to
 // a different map based on the value of the global $brand-theme variable.
 
+//* -------------------------------------------------------------------------- */
+// Vertical and Horizontal Spacing
+
+
+$theme-spacing-firefox: (
```
craigcook, line 13: Let's rename this `$theme-spacing` so it's not tied explicitly to the brand. 

In the near future we'll define a condensed spacing and probably give set it with a separate theme variable like we do with the condensed type scale, but for now just defining these units gives us a good start.

craigcook (changes requested):
LGTM, let's just rename the map so it's distinct from the brand theme.

src/assets/sass/protocol/includes/_themes.scss
```diff
@@ -6,6 +6,34 @@
...
+    v-grid-xs: $layout-md,
+    h-grid-xs: $layout-xs,
+
+    // sm is the same as xs
+    v-grid-sm: $layout-md,
+    h-grid-sm: $layout-xs,
+
+    // md
+    v-grid-md: $layout-md,
+    h-grid-md: $layout-md,
```
craigcook, line 24: Just noticed: Should this be `$layout-lg` (64px)? Referring to https://github.com/mozilla/protocol/issues/345#issuecomment-584856741

craigcook (changes requested):

src/assets/sass/protocol/includes/_themes.scss
stephaniehobson: Oh, yes.

stephaniehobson:
Updated.

craigcook (approved):

PR #670, opened 2021-03-20
Logos

### LINKED ISSUE #663: Add VPN logo support
The VPN site is joining the Protocol family and we should support adding VPN logos to things.

### LINKED ISSUE #665: Backport Logo Component from bedrock
More than once we have used image replacement techniques to add logos to things. We should add a way to do that in Protocol.

Here is one implementation:
https://github.com/mozilla/bedrock/blob/53741200257a192422e1a11ec4defdf1a50c84c4/media/css/firefox/new/desktop/download.scss#L297

For Contenful purposes it would be useful if we could nest the component inside a semantic wrapper like: `<h2><div class="c-logo">Firefox Browser</div></h2>` instead of applying the class to the parent heading. Open for discussion though.

stephaniehobson:
## Description

- Update to protocol-assets v4
    - Adds VPN and Focus logos
    - Updates Mozilla and Pocket logos
- Update hero and callout classes to support new products
- Add logo and wordmark components

---

### Issue

Fix #663
Fix #665

### Testing

New components:
http://localhost:3000/patterns/atoms/logo.html
http://localhost:3000/demos/logo.html

Try new product classes on:
http://localhost:3000/patterns/organisms/hero.html
http://localhost:3000/patterns/organisms/call-out.html

### DIFF
--- a/src/patterns/atoms/logo/logo.hbs
+++ b/src/patterns/atoms/logo/logo.hbs
@@ -2,13 +2,13 @@
 name: Logos
 description: |
   1. Add the CSS files to your bundle. You will need both the file for logo and for the product logo you want.
-  2. Add the component class `mzp-c-logo` the size class (eg: `mzp-t-md`) and the product class (eg: `mzp-t-product-firefox`) to your markup
+  2. Add the component class `mzp-c-logo` the size class (eg: `mzp-t-logo-md`) and the product class (eg: `mzp-t-product-firefox`) to your markup
      - For vertical spacing put the logo classes on or inside a heading or paragraph.
 order: 2
 notes: |
 tips: |
   - Logos need both a size and product theme to display.
-    - Size classes are: `mzp-t-xs`, `mzp-t-sm`, `mzp-t-md`, `mzp-t-lg`, `mzp-t-xl`.
+    - Size classes are: `mzp-t-logo-xs`, `mzp-t-logo-sm`, `mzp-t-logo-md`, `mzp-t-logo-lg`, `mzp-t-logo-xl`.
     - Product classes are: `mzp-t-product-family`, `mzp-t-product-firefox`, `mzp-t-product-beta`, `mzp-t-product-developer`, `mzp-t-product-nightly`, `mzp-t-product-focus`, `mzp-t-product-lockwise`, `mzp-t-product-monitor`, `mzp-t-product-mozilla`, `mzp-t-product-vpn`, `mzp-t-product-pocket`.
 ---
 
--- a/src/patterns/atoms/logo/logo.hbs
+++ b/src/patterns/atoms/logo/logo.hbs
@@ -1,0 +1,21 @@
+---
+name: Logos
+description: |
+  This component uses an image replacement technique for better SEO. You can, of course, still use a regular img element to insert a logo if it fits your use case better. Don't forget the alt text.
+
+  1. Add the CSS files to your bundle. You will need both the file for logo and for the product logo you want.
+  2. Add the component class `mzp-c-logo` the size class (eg: `mzp-t-md`) and the product class (eg: `mzp-t-product-firefox`) to your page
+     - For vertical spacing put the logo classes on or inside a heading or paragraph.
+order: 1
+notes: |
+tips: |
+  - Logos need both a size and product theme to display.
+    - Size classes are: `mzp-t-xs`, `mzp-t-sm`, `mzp-t-md`, `mzp-t-lg`, `mzp-t-xl`.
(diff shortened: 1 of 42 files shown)


src/patterns/atoms/logo/logo.hbs
```diff
@@ -0,0 +1,21 @@
...
+  This component uses an image replacement technique for better SEO. You can, of course, still use a regular img element to insert a logo if it fits your use case better. Don't forget the alt text.
+
+  1. Add the CSS files to your bundle. You will need both the file for logo and for the product logo you want.
+  2. Add the component class `mzp-c-logo` the size class (eg: `mzp-t-md`) and the product class (eg: `mzp-t-product-firefox`) to your page
+     - For vertical spacing put the logo classes on or inside a heading or paragraph.
+order: 1
+notes: |
+tips: |
+  - Logos need both a size and product theme to display.
+    - Size classes are: `mzp-t-xs`, `mzp-t-sm`, `mzp-t-md`, `mzp-t-lg`, `mzp-t-xl`.
```
craigcook, line 13: I've been using theme/variant classes that reference the component, such as `mzp-t-content-md`, I guess just to make it clearer what it's applied to but it might also avoid potential conflicts if we end up with something that gets two different theme classes. 

So these could be `mzp-t-logo-md` (or `mzp-t-wordmark-md`). The benefit of clarity might be worth the cost of verbosity.

craigcook (changes requested):

src/patterns/atoms/logo/logo.hbs
stephaniehobson: Other review comments fixed. Still thinking about how I feel about this one. I was following existing conventions from stuff like buttons but that's not a reason not to do something better. I'm currently avoiding conflicts by requiring them to be used with the component class but that doesn't mean explicit isn't better. I like the shortness and predictability of using `mzp-t-sm` everywhere you want a small thing but with gzip I'm not sure there's any performance benefits. Anyway, need to sleep on this again I guess.

craigcook: Yeah, I've gone back and forth too. I ended up doing longer classes on the content container because when it was just "mzp-t-md" that didn't feel descriptive enough, at least in that case... like "what's medium here?" since it was a layout wrapper containing lots of things, and not a single component. The more explicit name also matched up with the tokens so it felt right. But on buttons it's on a small, specific component so it's more obvious what the size applies to... and the same is true here.

But in general I think maybe explicit is better than implicit, as long as we can still keep it pretty short. But does that mean we should change the buttons to all be `mzp-t-button-md`? Maybe so... 

Naming things is hard.

stephaniehobson: Updated the names.

src/patterns/atoms/logo/logo.hbs
```diff
@@ -0,0 +1,18 @@
...
+name: Logos
+description: |
+  1. Add the CSS files to your bundle. You will need both the file for logo and for the product logo you want.
+  2. Add the component class `mzp-c-logo` the size class (eg: `mzp-t-md`) and the product class (eg: `mzp-t-product-firefox`) to your markup
+     - For vertical spacing put the logo classes on or inside a heading or paragraph.
+order: 2
+notes: |
+tips: |
+  - Logos need both a size and product theme to display.
+    - Size classes are: `mzp-t-xs`, `mzp-t-sm`, `mzp-t-md`, `mzp-t-lg`, `mzp-t-xl`.
```
craigcook, line 11: Need to update the docs with the new name.

src/patterns/atoms/logo/wordmark.hbs
```diff
@@ -0,0 +1,20 @@
...
+  Wordmarks work a lot like the logos except they change when nested under a `.mzp-t-dark` class.
+
+  1. Add the CSS files to your bundle. You will need both the file for wordmark and for the product wordmark you want.
+  2. Add the component class `mzp-c-wordmark` the size class (eg: `mzp-t-md`) and the product class (eg: `mzp-t-product-firefox`) to your markup
+     - For vertical spacing put the wordmark classes on or inside a heading or paragraph.
+order: 3
+notes: |
+tips: |
+  - Wordmarks need both a size and product theme to display.
+    - Size classes are: `mzp-t-xs`, `mzp-t-sm`, `mzp-t-md`, `mzp-t-lg`, `mzp-t-xl`.
```
craigcook, line 13: Here too.

craigcook (approved):
Just need to update the size classes in the docs. r+wc 🌳

Issue #628, opened 2020-07-20
Add mzp-t-dark to fundamentals/themes


amychurchwell:
## Description

Add an example of using `mzp-t-dark` to fundamentals/themes

craigcook:
That page is changing a lot in https://github.com/mozilla/protocol/pull/609. 

We don't have a really fleshed out dark theme and I want to work on that soon. In the interim I can try to add some documentation of the rudimentary dark theme we do have.

amychurchwell:
@craigcook We have it on a per-component basis, which I think is worth calling out even if it's not a body tag theme yet.

rochisha0:
Can I work on this @amychurchwell ?

craigcook:
> Can I work on this @amychurchwell ?

Sorry, this is already being addressed in https://github.com/mozilla/protocol/pull/609, I should have assigned myself.

stephaniehobson:
Dark themes are documented on a per-component basis (we don't have dark brand themes... yet).

PR #676, opened 2021-03-30
Add theme class for alt background color [no bug]

craigcook:
## Description

This adds a common theme class for applying an alternative background color: `.mzp-t-background-alt`
Also standardizes `.mzp-t-dark` class to use the theme variable for inverse backgrounds.

I haven't really documented it because I'm not sure where that documentation should go... we probably need to think about that. I'd suggest a new doc page for these kinds of utility and theme classes. There's [an older issue about properly documenting the "dark" theme](#628 ), that would go in the same place. I can work on better docs but I don't want to hold up getting this class added.

### Testing

http://localhost:3000/demos/hero.html uses the alt background class on a few examples

### DIFF
--- a/src/assets/sass/protocol/components/_call-out.scss
+++ b/src/assets/sass/protocol/components/_call-out.scss
@@ -12,7 +12,7 @@
     text-align: center;
 
     &.mzp-t-dark {
-        background-color: get-theme('background-color-inverse');
+        background-color: get-theme('background-color-alt-inverse');
         color: get-theme('body-text-color-inverse');
 
         .mzp-c-call-out-desc {
@@ -97,14 +97,14 @@
 // Compact Call Out component
 
 .mzp-c-call-out-compact {
-    background-color: $color-marketing-gray-20;
+    background-color: get-theme('background-color-alt');
 
     &.mzp-t-dark {
-        background-color: $color-marketing-gray-80;
-        color: $color-white;
+        background-color: get-theme('background-color-alt-inverse');
+        color: get-theme('body-text-color-inverse');
 
         .mzp-c-call-out-desc {
-            color: $color-marketing-gray-30;
+            color: get-theme('body-text-color-inverse');
         }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Features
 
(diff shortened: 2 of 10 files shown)


craigcook:
I wonder if the default alt color should be a shade darker. It's pretty subtle, and I don't recall where I took that color from originally (there was some page that used it). Maybe we should seek an opinion from The Studio?

src/assets/sass/protocol/components/_call-out.scss
```diff
@@ -8,11 +8,12 @@
 // Call Out component
 
 .mzp-c-call-out {
-    background-color: get-theme('background-color-alt');
+    background-color: $color-marketing-gray-20;
```
stephaniehobson, line 11: This one did not end up with a theme variable.

stephaniehobson: I think it would also be okay to standardize on the lighter background colour.

src/assets/sass/protocol/components/_call-out.scss
```diff
@@ -11,8 +11,9 @@
     background-color: get-theme('background-color-alt');
     text-align: center;
 
-    &.mzp-t-dark {
-        background-color: get-theme('background-color-inverse');
+    &.mzp-t-dark,
+    .mzp-t-dark & {
```
stephaniehobson, line 15: I think we had talked about using `&.mzp-t-dark` as the standard way to theme components.

stephaniehobson (approved):
R+ 🟪  One possible change to a declaration but I'm not bothered.

stephaniehobson:
Oh, docs wise maybe add something to the themes page?

src/assets/sass/protocol/components/_call-out.scss
craigcook: Doh, this was a leftover from testing. Good catch.

src/assets/sass/protocol/components/_call-out.scss
craigcook: I did start on adding an accent color to themes but don't want to pile more into this PR. We can think more about accent colors another time.

craigcook:
I added a bit to the themes page but I'm starting a draft of a more extensive "Using Protocol" doc and I think that content can eventually go there.

PR #680, opened 2021-04-09
Bump version to 14.0.1

craigcook:
## Description

This includes a bug fix in the main _lib.scss and also fixes the logo in the doc site nav (which I missed because I still had old assets on my local 🤦). I'll be more careful when publishing this.


stephaniehobson (approved):
R+ 🧷

PR #688, opened 2021-05-11
Set global Mzp namespace to default to window as root (Fixes #687)

### LINKED ISSUE #687: Set `window` as default global property for JS namespace
## Description

Protocol JS components use an `Mzp` namespace (e.g. Mzp.Navigation, Mzp.Newsletter). In the browser, the default root property is `window`, so these would be accessible via `window.Mzp.Navigation`, `window.Mzp.Newsletter` etc.

Many JS bundlers (such as Webpack, Parcel etc) default to wrapping scripts inside a closure to avoid leaking global variables. This is generally a good practise, however in the case of Protocol the `Mzp` namespace is intended to be global (at least in terms for how it is built today). The result is that when you bundle together several Protocol components into a single script, `window.Mzp` is undefined since `Mzp` is wrapped inside a closure, making it hard to initiate components from outside where they are defined.

I believe this can be fixed via a small change to the way we initiate `Mzp`:

Before:

```javascript
// create namespace
if (typeof Mzp === 'undefined') { 
    var Mzp = {};
}
```

After:

```javascript
// create namespace
if (typeof window.Mzp === 'undefined') {
    window.Mzp = {};
}
```

alexgibson:
## Description

Ensures JS components are always accessible via `window.Mzp` namespace.

~- [ ] I have documented this change in the design system.~

### Issue

#687

### Testing

- [ ] Components should still be accessible via `window.Mzp` as before by default.
- [ ] Check I didn't miss anywhere else?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# HEAD
+
+## Bug Fixes
+
+* **js:** Set global `Mzp` namespace to default to `window` as root (#687).
+
 # 14.0.3
 
 ## Bug Fixes
--- a/src/assets/js/protocol/protocol-details.js
+++ b/src/assets/js/protocol/protocol-details.js
@@ -4,8 +4,8 @@
 
 
 // create namespace
-if (typeof Mzp === 'undefined') { // eslint-disable-line block-scoped-var
-    var Mzp = {};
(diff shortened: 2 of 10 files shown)


craigcook (approved):
🌐

PR #703, opened 2021-08-09
Enhancements/fixes for Split component [fix #702]

### LINKED ISSUE #702: Enhancements/fixes for Split component
As we've been implementing Split on more pages on mozilla.org we've identified a few issues to address:

1. The split container is using the border-box model, which makes the Split content significantly narrower (in large viewports) than other content containers that use the default content-box.
image

2. Split doesn't switch to a two column layout until the large (desktop) breakpoint, where the older Hero and Feature Card components switch to two columns at the medium (tablet) breakpoint.
3. The media in a Split has a default max-width in small viewports, instead of expanding to the available width. This often doesn't look good in medium-to-large sized viewports (especially combined with 2 above). 
image

## Proposed changes
1. Remove the border-box declaration, let Split follow the default content-box model as other content containers (we'll need to test thoroughly to make sure this doesn't cause other issues with Split layout).
2. Trigger the two column layout at the medium breakpoint (`$mq-md`). This won't look good in the narrow-body variant, so perhaps narrow-body should be 50%/50% at medium and only switch to 33%/66% at large.
3. Remove the max-width for media, let it expand to fit the container.

craigcook:
## Description
* Use content-box for Split container.
* Trigger column layout at md breakpoint.
* Remove max-width from media in small viewports.

### Issue
#702 

### Testing
http://localhost:3000/demos/split.html
http://localhost:3000/demos/split-visual.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ## Bug Fixes
 
+* **css:** Enhancements/fixes to Split component (#702)
 * **css:** Let Picto retain its bottom margin in multi-column layouts in small viewports (#699)
 * **css:** Override styling native summary element when it’s polyfilled in IE11 (#658)
 * **js:** Set global `Mzp` namespace to default to `window` as root (#687).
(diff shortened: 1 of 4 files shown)


alexgibson (approved):
Needs a rebase, but r+ on the changes 💯

Issue #712, opened 2021-08-22
Add some default spacing below the logo/wordmark component


craigcook:
In most cases a logo or wordmark appears above a headline or some other content and we've found that we always need to add a bottom margin. Since the typical use-case for the component seems to need spacing more often than not, we should add a margin to the component. We can still override it on the rare occasion we need a logo/wordmark _without_ spacing, but most of the time it wouldn't need any added styling.

PR #713, opened 2021-08-24
Split updates, logo margins, 3xs title class

craigcook:
## Description

* Added content width classes to Split component. This will help it play nicer when combined with other widths on the same page.
* Add bottom margins to logos and wordmarks [see #712]
* Added an `mzp-u-title-3xs` utility class. We revised the type scale after these classes were created and added another heading size, but didn't add a class for the new one.
* Misc doc updates. Rewrote Split docs for better clarity and easier reference.

### Issue
#712 - add default bottom margins on logos

### Testing
http://localhost:3000/demos/split.html
http://localhost:3000/patterns/atoms/logo.html

### DIFF
--- a/src/assets/sass/protocol/base/utilities/_titles.scss
+++ b/src/assets/sass/protocol/base/utilities/_titles.scss
@@ -42,3 +42,8 @@
     @include text-title-2xs;
     font-family: get-theme('title-font-family');
 }
+
+.mzp-u-title-3xs {
+    @include text-title-3xs;
+    font-family: get-theme('title-font-family');
+}
--- a/src/assets/sass/protocol/components/_split.scss
+++ b/src/assets/sass/protocol/components/_split.scss
@@ -17,15 +17,15 @@
 
     @media #{$mq-md} {
         padding: get-theme('v-grid-md') 0;
+
+        &.mzp-t-split-nospace {
+            padding-top: 0;
+            padding-bottom: 0;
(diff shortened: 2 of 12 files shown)


src/assets/sass/protocol/components/_split.scss
```diff
@@ -17,16 +17,16 @@
 
     @media #{$mq-md} {
         padding: get-theme('v-grid-md') 0;
+
+        &.mzp-t-split-nospace {
```
craigcook, line 21: Moved this into the media query so it still gets default spacing on small screens (since media can't pop out on mobile anyway).

src/assets/sass/protocol/base/utilities/_titles.scss
```diff
@@ -42,3 +42,8 @@
     @include text-title-2xs;
     font-family: get-theme('title-font-family');
 }
+
+.mzp-u-title-3xs {
+    @include text-title-3xs;
+    font-family: get-theme('title-font-family');
+}
```
nathan-barrett, line 49: Nice!

maureenlholland (approved):
I really love the layout visualisations. I didn't realize it was different from demos when I first looked at the docs. Would be cool to find a way to integrate those visualisations into the main docs (similar to the rtl toggle).

Optional change: include a video example.

Happy to see this merge as is though. r+ 👏

Issue #714, opened 2021-08-26
Add dark-theme support to Card component


alexgibson:
### Please describe the use case(s) for this component

Enable use of the Card component on a dark themed page / section.

### Please describe the change you wish to see

Currently the Card component comes with a solid white background color. Even when overriding this with `background: inherit;`, neither the heading or description support dark-theme color styling.

### Please describe why these changes need to be made

So we can use Card components on pages that support both light and dark theme styles, such as /whatsnew on www.mozilla.org.

### Please provide any designs or prototypes of the proposed changes

Default styling (incorrect):

<img src="https://user-images.githubusercontent.com/400117/130962414-fb313c27-b5a3-4257-aedf-52ba7c336310.png" alt="image" width="500">

See below for how it should look on a dark-themed page.

### Please describe where this component is currently in use

The Card component is in use on many pages on www.mozilla.org, but none with a dark-themed background (until now).

### Please describe what this component should do in the following conditions

* When set in front of a dark background

<img src="https://user-images.githubusercontent.com/400117/130962723-2af5e79a-db30-4b3d-977b-987343aaef17.png" alt="image" width="500">

* When set in front of a light background

<img src="https://user-images.githubusercontent.com/400117/130962890-0824a954-08a6-4c17-9d99-2d746c993a3c.png" alt="image" width="500">

* When viewed in a mobile/responsive viewport

N/A

### Is the development of this component a blocking dependency for other work? Please explain if so

No (I can manage with an override for now), but we should still fix this.

PR #715, opened 2021-09-03
had rules to card with mzp-t-dark class

nathan-barrett:
## Description
Added some rules when adding the `.mzp-t-dark` class to enable dark mode styled cards. I used the same themes as in other areas of protocol, but let me know if anything should change. 

Added some rules to override the links and `p` tags in the card component so the color would be visible for dark mode. Since dark mode isn't listed outside of the themes section, I wasn't sure how it should be documented. 

Screen Shot 2021-09-07 at 3 48 22 PM

### Issue
#714

### Testing
Wasn't sure how we would want to add this card _variant_ to the documentation since dark mode is only referenced in the themes section of the docs. If you build this locally, you can just add the `.mzp-t-dark` class to any card to it's container. I created the rules so it would have to be on whichever element has the `mzp-c-card` class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # HEAD
 
 ## Features
-* **css:** Added overidding rules to the card component to enable dark mode with `.mzp-t-dark` class.
+* **css:** Add overidding rules to the card component to enable dark mode with `.mzp-t-dark` class.
 * **css:** Add content width classes to Split component.
 * **css:** Add default bottom margins to Logo and Wordmark components (#712)
 * **css:** Add `.mzp-u-title-3xs` utility class.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # HEAD
 
 ## Features
-
+* **css:** Added overidding rules to the card component to enable dark mode with `.mzp-t-dark` class.
 * **css:** Add content width classes to Split component.
 * **css:** Add default bottom margins to Logo and Wordmark components (#712)
 * **css:** Add `.mzp-u-title-3xs` utility class.
--- a/src/assets/sass/protocol/components/_card.scss
+++ b/src/assets/sass/protocol/components/_card.scss
@@ -147,6 +147,10 @@
     }
 
     .mzp-c-card-block-link {
+        &:hover, &:active, &:focus {
+            box-shadow: 0 0 0 4px $color-marketing-gray-80;
+        }
+
         &:link, &:visited {
         color: get-theme('body-text-color-inverse');
 
--- a/src/assets/sass/protocol/components/_card.scss
+++ b/src/assets/sass/protocol/components/_card.scss
@@ -25,6 +25,7 @@
         @include aspect-ratio(3, 2);
     }
 
+
     .mzp-c-card-media-wrapper {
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # HEAD
 
 ## Features
-
+* **css:** Added overidding rules to the card component to enable dark mode with `.mzp-t-dark` class.
```
nathan-barrett, line 4: Can change this to `Add` before merge

nathan-barrett:
@alexgibson let me know what you think - the color followed the `color-inverse` theme already in place in protocol, but I wasn't sure if that's the one we wanted.

src/assets/sass/protocol/components/_card.scss
```diff
@@ -134,6 +134,24 @@
     }
 }
 
+//* -------------------------------------------------------------------------- */
+// Dark mode styling
+
+.mzp-c-card.mzp-t-dark {
```
craigcook, line 140: We should also add a selector that lets cards in a dark container take on the theme:
```suggestion
.mzp-c-card.mzp-t-dark,
.mzp-t-dark .mzp-c-card {
```

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # HEAD
 
 ## Features
-
+* **css:** Add overidding rules to the card component to enable dark mode with `.mzp-t-dark` class.
```
craigcook, line 4: Typo: overriding

craigcook (changes requested):
We should mention the dark theme in the docs on the component page, and maybe include an example on one of the demo pages.

src/pages/demos/card-layout.hbs
```diff
@@ -85,5 +85,21 @@ styles:
   </div>
 </div>
 
+<div class="mzp-l-content">
```
craigcook, line 88: I think the demo will be more effective if we make the container dark:
```suggestion
<div class="mzp-l-content mzp-t-dark">
```

src/pages/demos/card-layout.hbs
```diff
@@ -85,5 +85,21 @@ styles:
...
+  <header>
+    <h2>2 Card Layout - dark mode</h2>
+    <ul class="mzp-u-list-styled">
+      <li>To use dark mode either place the `.mzp-t-dark` class on the element that contains the `.mzp-c-card` class or place the card
+      within a container that is using the dark mode class.</li>
+    </ul>
+  </header>
+
+  <div class="mzp-l-card-half">
+    {{#embed "patterns.molecules.card.small-card" class="mzp-t-dark"}}{{/embed}}
```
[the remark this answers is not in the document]
craigcook: Actually, I just noticed the medium card include also adds a tag with an icon, but we don't have any light icons in Protocol (we need to update all the icons anyway, but that's a whole other thing) so it doesn't look great as a demo. 

image

So maybe we stick with the small molecule but pass it the medium class explicitly (which is all the medium card include does anyway):

```suggestion
    {{#embed "patterns.molecules.card.small-card" class="mzp-c-card-medium mzp-has-aspect-3-2"}}{{/embed}}
```

craigcook (changes requested):

craigcook (approved):

PR #719, opened 2021-09-13
Bump version to 14.1.0

craigcook:


alexgibson (changes requested):
I think you also need to bump the version number in https://github.com/mozilla/protocol/blob/main/src/assets/package/package.json

craigcook:
> I think you also need to bump the version number in https://github.com/mozilla/protocol/blob/main/src/assets/package/package.json

D'oh, I knew I was forgetting something. Updated.

alexgibson:
Sorry I missed this, there's also a version numnber in the readme that looks like it needs updating too https://github.com/mozilla/protocol/blob/main/src/assets/package/README.md

alexgibson (approved):
Looks good! 🩹

PR #720, opened 2021-09-14
Update Node version in build.sh

craigcook:
## Description

Recent deployments have been failing because our versions of some dependencies no longer work in Node 10, which is what the build script is using. This updates the build to Node 14 which should support the other dependencies.

Recent failure: https://gitlab.com/mozmeao/protocol/-/jobs/1589827585

`Error: yargs supports a minimum Node.js version of 12. Read our version support policy: https://github.com/yargs/yargs#supported-nodejs-versions`


pmac (approved):
💯

PR #721, opened 2021-09-14
Rename master to main in gitlab-ci.yml

craigcook:
Stage deployments were still watching for changes in a branch that no longer exists. This should fix it.


pmac (approved):
👍

Issue #723, opened 2021-09-15
Handle protocol-assets PNG removals


maureenlholland:
A [breaking change](https://github.com/mozilla/protocol-assets/issues/65) is coming in protocol-assets.

⚠️ This will likely be a breaking change for protocol as well.

Success criteria:
- Logo and wordmark mixins use SVG with expected sizing based on existing size classes
- No unexpected breaking images with protocol-assets branch

PR #724, opened 2021-09-15
Use SVG instead of PNG for logos and wordmarks [#723]

maureenlholland:
## Description

We are working towards a smaller protocol-assets dependency size. This will involve removing PNG assets, which will break our current logo and wordmark mixins. We want to preserve the use of these mixins, but return SVG assets instead of PNG.

~⚠️ Cannot merge until Focus logo is available in SVG with next protocol-assets version~ Merged! All clear

### Issue

https://github.com/mozilla/protocol/issues/723

### Testing
TBA


maureenlholland:
Closed in favour of: https://github.com/mozilla/protocol/pull/728

PR #726, opened 2021-09-20
Update dependencies

### LINKED ISSUE #722: Update Sass, replace slash divisions
Protocol is currently on gulp-sass 4.1.1. As of gulp-sass 5.0.0, it no longer includes its own Sass compiler and requires an additional package, either sass or node-sass. While testing it with the latest Sass package I learned that [using `/` as a division operator is being deprecated](https://sass-lang.com/documentation/breaking-changes/slash-div). 

We should weed out all our uses of `/` for division in our Sass/SCSS and change them to the new `math.div()` function. Sass provides a migration tool that should help automate this but we'll want to take care to make sure nothing breaks along the way.

The current version of node-sass doesn't seem to be using the latest Sass yet so if we use that as our compiler it won't show any warnings. But it'll become an issue in the near future so we might as well get ahead of it.

craigcook:
## Description
Updates various dependencies, should address some security alerts (at least until we can migrate off Drizzle).

This also fixes #722 by adding dart-sass and replacing all our `/` division operators (which was fewer than I thought).

### DIFF
--- a/src/assets/sass/protocol/components/_footer.scss
+++ b/src/assets/sass/protocol/components/_footer.scss
@@ -2,6 +2,8 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+@use "sass:math";
+
 @import '../includes/lib';
 
 //* -------------------------------------------------------------------------- */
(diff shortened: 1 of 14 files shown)


maureenlholland (changes requested):
Looks good! Needs a changelog note, then r+

maureenlholland:
oh hang on, poor review on my part, there might be a few more `/` operators lurking
Screen Shot 2021-09-21 at 4 29 33 PM

craigcook:
Huh. The migration tool didn't find those and they don't raise deprecation warnings, so evidently Sass doesn't recognize when we do math with tokens. I'll fix them and see if I can find any more.

src/assets/sass/protocol/components/_footer.scss
```diff
@@ -99,36 +101,37 @@
...
     @for $cols from 2 through 6 {
-        $width: 100% / $cols;
+        $width: math.div(100%, $cols); // each width is the total width divided by number of columns
+        $gutter: math.div($layout-md, $cols); // each gutter is the base gutter divided by number of columns
+
         // http://lea.verou.me/2011/01/styling-children-based-on-their-number-with-css3/
         .mzp-c-footer-section:first-child:nth-last-child(n+#{$cols}),
         .mzp-c-footer-section:first-child:nth-last-child(n+#{$cols}) ~ .mzp-c-footer-section:nth-child(1n-1) {
-            width: calc(#{$width} - (#{$layout-md} - (#{$layout-md} / #{$cols})));
+            width: calc($width - ($layout-md - $gutter));
```
maureenlholland, line 134: Lovely upgrade for sass and readability!

maureenlholland (approved):
r++ 💠

PR #730, opened 2021-09-30
Create new breadcrumb component

nathan-barrett:
## Description
Created a new breadcrumb component which will be initially used in theVPN Resource Center, but will most likely be used in other areas on mozilla.org. 

### Testing

Start your local server and navigate to http://localhost:3000/patterns/molecules/breadcrumb.html# and compare to current Figma [design](https://www.figma.com/file/xo5JhiQIlpGWkb9lgh0ccZ/VPN-Resource-Guide?node-id=16%3A256)
Screen Shot 2021-09-29 at 5 16 55 PM
Short text on mobile sizing:
Screen Shot 2021-09-29 at 5 17 05 PM

### DIFF
--- a/src/assets/sass/protocol/components/_breadcrumb.scss
+++ b/src/assets/sass/protocol/components/_breadcrumb.scss
@@ -7,30 +7,30 @@
 .mzp-c-breadcrumb {
     background-color: #f3f3f3;
     margin-bottom: 0;
-    padding: 8px 18px;
+    padding: $spacing-sm $spacing-md;
 
     .mzp-c-breadcrumb-item {
-        display: inline;
+        font-weight: bold;
+        &:before {
+            @include bidi(((content, "\2192", "\2190"),));
+            font-weight: normal;
+            margin: 0 0.25em;
+        }
+    }
+
+    .mzp-c-breadcrumb-item, .mzp-c-breadcrumb-title {
+        display: inline-block;
         font-weight: 400;
 
         .mzp-c-breadcrumb-item-link {
             text-decoration: none;
-        }
-
-        &:not(.mzp-c-breadcrumb-title) {
-            font-weight: bold;
-            &:before {
-                content: ">";
-                font-weight: normal;
-                margin: 0 2px;
-            }
         }
     }
     &.mzp-t-dark {
         background-color: $color-black;
     }
     @media #{$mq-md} {
-        padding: 8px 80px;
+        padding: $spacing-sm $layout-xl - $spacing-md;
     }
 
 }
--- a/src/assets/sass/protocol/components/_breadcrumb.scss
+++ b/src/assets/sass/protocol/components/_breadcrumb.scss
@@ -5,24 +5,23 @@
 @import '../includes/lib';
 
 .mzp-c-breadcrumb {
-    background-color: $color-light-gray-20;
+    background-color: get-theme('background-color-alt');
     margin-bottom: 0;
     padding: $spacing-sm $spacing-md;
 
     .mzp-c-breadcrumb-item {
         display: inline-block;
+        font-weight: 400;
 
-        &:first-child {
-            font-weight: 400;
+        &:last-child {
+            font-weight: bold;
         }
 
         + .mzp-c-breadcrumb-item {
-            font-weight: bold;
-
             &:before {
-                @include bidi(((content, "\003E", "\003C"),));
+                @include bidi(((content, "\2192", "\2190"),));
                 font-weight: normal;
-                margin: 0 0.25em;
+                margin: 0 .25em;
             }
         }
 
@@ -31,10 +30,8 @@
         }
     }
 
-
-
     &.mzp-t-dark {
-        background-color: $color-black;
+        background-color: get-theme('background-color-alt');
     }
 
     @media #{$mq-md} {
@@ -44,5 +41,4 @@
     @media #{$mq-lg} {
         padding: $spacing-sm $layout-xl - $spacing-md;
     }
-
 }
--- a/src/assets/sass/protocol/components/_breadcrumb.scss
+++ b/src/assets/sass/protocol/components/_breadcrumb.scss
@@ -1,0 +1,36 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import '../includes/lib';
+
+.mzp-c-breadcrumb {
+    background-color: #f3f3f3;
+    margin-bottom: 0;
+    padding: 8px 18px;
+
+    .mzp-c-breadcrumb-item {
+        display: inline;
+        font-weight: 400;
+
+        .mzp-c-breadcrumb-item-link {
+            text-decoration: none;
+        }
+
+        &:not(.mzp-c-breadcrumb-title) {
+            font-weight: bold;
+            &:before {
+                content: ">";
+                font-weight: normal;
+                margin: 0 2px;
+            }
+        }
+    }
+    &.mzp-t-dark {
+        background-color: $color-black;
+    }
+    @media #{$mq-md} {
+        padding: 8px 80px;
+    }
+
+}
--- a/src/assets/sass/protocol/components/_breadcrumb.scss
+++ b/src/assets/sass/protocol/components/_breadcrumb.scss
@@ -6,39 +6,41 @@
 
 .mzp-c-breadcrumb {
     background-color: get-theme('background-color-alt');
-    margin-bottom: 0;
-    padding: $spacing-sm $spacing-md;
 
-    .mzp-c-breadcrumb-item {
-        display: inline-block;
-        font-weight: 400;
+    .mzp-c-breadcrumb-list {
+        margin-bottom: 0;
+        padding: $spacing-sm $spacing-md;
 
-        &:last-child {
-            font-weight: bold;
-        }
+        .mzp-c-breadcrumb-item {
+            display: inline-block;
 
-        + .mzp-c-breadcrumb-item {
-            &:before {
-                @include bidi(((content, "\2192", "\2190"),));
-                font-weight: normal;
-                margin: 0 .25em;
+            &:last-child {
+                font-weight: bold;
+            }
+
+            + .mzp-c-breadcrumb-item {
+                &:before {
+                    @include bidi(((content, "\2192", "\2190"),));
+                    font-weight: normal;
+                    margin: 0 .25em;
+                }
+            }
+
+            a {
+                text-decoration: none;
             }
         }
 
-        .mzp-c-breadcrumb-item-link {
-            text-decoration: none;
+        @media #{$mq-md} {
+            padding: $spacing-sm $layout-sm;
+        }
+
+        @media #{$mq-lg} {
+            padding: $spacing-sm $layout-xl - $spacing-md;
         }
     }
 
     &.mzp-t-dark {
-        background-color: get-theme('background-color-alt');
-    }
-
-    @media #{$mq-md} {
-        padding: $spacing-sm $layout-sm;
-    }
-
-    @media #{$mq-lg} {
-        padding: $spacing-sm $layout-xl - $spacing-md;
+        background-color: get-theme('background-color-alt-inverse');
     }
 }
--- a/src/assets/sass/protocol/components/_breadcrumb.scss
+++ b/src/assets/sass/protocol/components/_breadcrumb.scss
@@ -5,31 +5,43 @@
 @import '../includes/lib';
 
 .mzp-c-breadcrumb {
-    background-color: #f3f3f3;
+    background-color: $color-light-gray-20;
     margin-bottom: 0;
     padding: $spacing-sm $spacing-md;
 
     .mzp-c-breadcrumb-item {
-        font-weight: bold;
-        &:before {
-            @include bidi(((content, "\2192", "\2190"),));
-            font-weight: normal;
-            margin: 0 0.25em;
-        }
-    }
-
-    .mzp-c-breadcrumb-item, .mzp-c-breadcrumb-title {
         display: inline-block;
-        font-weight: 400;
+
+        &:first-child {
+            font-weight: 400;
+        }
+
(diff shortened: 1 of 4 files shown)


nathan-barrett:
There are still some small decisions / changes that need to be made before this would be ready to be merged. In addition, this is more of a first pass at a new protocol component and am happy to make changes to improve this.

src/assets/sass/protocol/components/_breadcrumb.scss
```diff
@@ -0,0 +1,36 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import '../includes/lib';
+
+.mzp-c-breadcrumb {
+    background-color: #f3f3f3;
```
craigcook, line 8: This should be a token, and preferably a theme property.

src/assets/sass/protocol/components/_breadcrumb.scss
```diff
@@ -0,0 +1,36 @@
...
+    margin-bottom: 0;
+    padding: 8px 18px;
```
craigcook, line 10: Use tokens for padding.

src/assets/sass/protocol/components/_breadcrumb.scss
```diff
@@ -0,0 +1,36 @@
...
+        .mzp-c-breadcrumb-item-link {
+            text-decoration: none;
+        }
+
+        &:not(.mzp-c-breadcrumb-title) {
+            font-weight: bold;
+            &:before {
+                content: ">";
+                font-weight: normal;
+                margin: 0 2px;
```
craigcook, line 25: Use tokens for margins, but our smallest spacing token is 4px. However, it might be better to use ems for this so it's proportional to the font size. Or maybe ch.

src/assets/sass/protocol/components/_breadcrumb.scss
```diff
@@ -0,0 +1,36 @@
...
+        font-weight: 400;
+
+        .mzp-c-breadcrumb-item-link {
+            text-decoration: none;
+        }
+
+        &:not(.mzp-c-breadcrumb-title) {
+            font-weight: bold;
+            &:before {
+                content: ">";
```
craigcook, line 23: We probably shouldn't use the ">" character because it isn't actually an arrow, it already has a meaning (greater than). We can find some other unicode arrow character or maybe an SVG icon.

We also need to accommodate RTL and reverse the direction.

src/assets/sass/protocol/components/_breadcrumb.scss
```diff
@@ -0,0 +1,36 @@
...
+                font-weight: normal;
+                margin: 0 2px;
+            }
+        }
+    }
+    &.mzp-t-dark {
+        background-color: $color-black;
+    }
+    @media #{$mq-md} {
+        padding: 8px 80px;
```
craigcook, line 33: Tokens for spacing.

src/assets/sass/protocol/components/_breadcrumb.scss
```diff
@@ -0,0 +1,36 @@
...
+
+    .mzp-c-breadcrumb-item {
+        display: inline;
+        font-weight: 400;
+
+        .mzp-c-breadcrumb-item-link {
+            text-decoration: none;
+        }
+
+        &:not(.mzp-c-breadcrumb-title) {
```
craigcook, line 20: Support for `:not` is really recent, we should probably find a more backwards-compatible method.

craigcook (changes requested):
I haven't done any real testing yet, just adding a bunch of drive-by comments on a first look through the code.

src/assets/sass/protocol/components/_breadcrumb.scss
nathan-barrett: https://developer.mozilla.org/en-US/docs/Web/CSS/:not seems like its the `selector list` that isn't well supported (`:not(div, .my-class)`) when its just one selector it seems like its supported back to IE 9. Should we still find something else to use?

src/assets/sass/protocol/components/_breadcrumb.scss
nathan-barrett: Ok, good call out

src/assets/sass/protocol/components/_breadcrumb.scss
craigcook: Ah, right you are. When I checked caniuse I was looking at the entry for multi-selector, but we should be fine with a single selector. My mistake 👍

src/patterns/molecules/breadcrumb/breadcrumb.hbs
```diff
@@ -0,0 +1,20 @@
+---
+name: Breadcrumb
+description: A type of navigation menu used to provide a list of links based on current page within the site heirarchy, allowing users to visit previous pages from the current page. Current implementation is to use this component as a header.
```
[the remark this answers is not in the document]
maureenlholland: "Current implementation is to use this component as a header" - I'm not following here, does this mean it should be at the top of the page as opposed to the bottom? Or it should be part of the header tag?

src/assets/sass/protocol/components/_breadcrumb.scss
```diff
@@ -0,0 +1,36 @@
...
+
+@import '../includes/lib';
+
+.mzp-c-breadcrumb {
+    background-color: #f3f3f3;
+    margin-bottom: 0;
+    padding: $spacing-sm $spacing-md;
+
+    .mzp-c-breadcrumb-item {
+        font-weight: bold;
```
maureenlholland, line 13: I think we're overriding the bold font type in the `-item`, `-title` selector below with 400 font weight

maureenlholland: I wonder too if it's worthwhile having separate classes for item and title or if we could use `:first-child` and `.mzp-c-breadcrumb-item + .mzp-c-breadcrumb-item` selectors to target the first and subsequent items in a way that could scale to multiple items in the breadcrumb.

maureenlholland (changes requested):
Great start! Only one actual requested changes for an overridden font-weight style and typo. The rest are just musings to take or leave.

Also, something we might want to ping Elise on. What to do on mobile when the breadcrumb is longer than one line? I don't know if we want a default wrap because it takes up such valuable space. Baymard seems to suggest a horizontal scroll is acceptable but I don't know if that's better. (Baymard doesn't link titles, but it's the How to Implement Long Breadcrumb Paths section: https://baymard.com/blog/implementing-mobile-hierarchy-breadcrumbs)
breadcrumb-mobile

src/assets/sass/protocol/components/_breadcrumb.scss
```diff
@@ -5,31 +5,43 @@
...
+                font-weight: normal;
+                margin: 0 0.25em;
+            }
+        }
 
         .mzp-c-breadcrumb-item-link {
             text-decoration: none;
         }
     }
+
```
maureenlholland, line 33: nit: remove extra empty lines

src/patterns/molecules/breadcrumb/breadcrumb.hbs
```diff
@@ -1,20 +1,21 @@
 ---
 name: Breadcrumb
-description: A type of navigation menu used to provide a list of links based on current page within the site heirarchy, allowing users to visit previous pages from the current page. Current implementation is to use this component as a header.
+description: A type of navigation menu used to provide a list of links based on current page within the site heirarchy, allowing users to visit previous pages from the current page.
```
maureenlholland, line 3:
```suggestion
description: A type of navigation menu used to provide a list of links based on current page within the site hierarchy, allowing users to visit previous pages from the current page.
```

maureenlholland (changes requested):
tiny nitpicks, and will need a changelog update - but more changelog conflict is coming with https://github.com/mozilla/protocol/pull/731

@craigcook do you think this or the protocol-assets dep update PR will merge first?

src/assets/sass/protocol/components/_breadcrumb.scss
```diff
@@ -0,0 +1,48 @@
...
+            font-weight: 400;
+        }
+
+        + .mzp-c-breadcrumb-item {
+            font-weight: bold;
+
+            &:before {
+                @include bidi(((content, "\003E", "\003C"),));
+                font-weight: normal;
+                margin: 0 0.25em;
```
craigcook, line 25: Nit: omit leading zeroes on decimals (doesn't matter in the grand scheme but we should be consistent).

craigcook (changes requested):

nathan-barrett:
This will probably need another rebase after 15.1 goes through, but I've double checked with Elise on the comments brought up in this PR and I _think_ have everything resolved. Let me know if anything else stands out. Would love to get this merged so we can use this when the VPN resource center goes live at end of Nov.

src/assets/sass/protocol/components/_breadcrumb.scss
```diff
@@ -0,0 +1,44 @@
...
+            }
+        }
+
+        .mzp-c-breadcrumb-item-link {
+            text-decoration: none;
+        }
+    }
+
+    &.mzp-t-dark {
+        background-color: get-theme('background-color-alt');
```
maureenlholland, line 34: Not sure we want `-alt` here as it's pretty light: https://github.com/mozilla/protocol/blob/main/src/assets/sass/protocol/includes/_themes.scss#L56

Screen Shot 2021-11-03 at 11 01 10 AM

Unless we also set color for `body-text-color-alt`? Or maybe change background to use `background-color-alt-inverse`?

maureenlholland (changes requested):
Thanks for checking in with Elise and following up on this! Would love to see it go into the wild with VRC 🍞 

A couple of nitty things and a question about dark theme colors. This is looking pretty good to me though.

src/assets/sass/protocol/components/_breadcrumb.scss
maureenlholland: Oh, yep, just saw a comment on this in a previous review. Prefer `background-color-alt-inverse`

craigcook:
I just came across this: https://html.spec.whatwg.org/multipage/semantics-other.html#rel-up which sent me down a short rabbit-hole, and I quickly found https://css-tricks.com/markup-for-breadcrumbs/ and then https://schema.org/BreadcrumbList

So in short, it seems like a paragraph with links may be preferable to a list. An ordered list does at least clearly indicate a sequence, but it doens't indicate hierarchy: the links are in a specific order but there's no parent<>child relationship. 

Just putting links in order without marking them up as a list already establishes the sequence without the added implication that they're all at the same level. Probably the best way to establish a hierarchy of links would be a series of nested lists, but that seems really clunky and could create more problems than it solves.

So then there's the idea of adding some more attributes to better describe the relationship. There was a proposed `rel="up"` attribute that seemed promising until it was axed. There's a breadcrumb schema at Schema.org but it's also clunky and seems like it helps search engines more than it helps humans.

So I'm a little torn on whether we should stick with an ordered list or just do a paragraph with inline links, and I could go either way.

But this little bit of research *has* convinced me that we should require breadcrumbs to be in a `nav` after all, since they are absolutely navigation and always will be. So I suggest `<nav class="mzp-c-breadcrumb">` as the component container.

src/patterns/molecules/breadcrumb/breadcrumb.hbs
```diff
@@ -0,0 +1,21 @@
+---
+name: Breadcrumb
+description: A type of navigation menu used to provide a list of links based on current page within the site hierarchy, allowing users to visit previous pages from the current page.
+order: 1
+notes: |
+    - Use the `.mzp-t-dark` options to use dark mode.
+    - Use a `<ol>` to use semantic HTML and structure.
+    - Using the `nav` element is optional, but reccomended.
+---
+<nav aria-label="breadcrumb">
```
craigcook, line 10: I also [commented elsewhere](https://github.com/mozilla/protocol/pull/730#issuecomment-961258254) but repeating here so it's not lost: after a little research I think we should indeed require breadcrumbs to be in a `nav` and make this `<nav class="mzp-c-breadcrumb" aria-label="breadcrumb">`

src/patterns/molecules/breadcrumb/breadcrumb.hbs
```diff
@@ -0,0 +1,21 @@
...
+  <ol class="mzp-c-breadcrumb">
```
craigcook, line 11: With the main component class moving to the `nav` I think this would become `mzp-c-breadcrumb-list`

src/patterns/molecules/breadcrumb/breadcrumb.hbs
```diff
@@ -0,0 +1,21 @@
...
+    <li class="mzp-c-breadcrumb-item">
+      <a class="mzp-c-breadcrumb-item-link" href="#">Parent page</a>
```
craigcook, line 13: I wonder if we really need a class on each link. Maybe any link in an `mzp-c-breadcrumb-item` gives us enough description and specificity?

craigcook (changes requested):

maureenlholland:
I'd stick with an ordered list for now, that seems to be the accepted accessible pattern: https://github.com/scottaohara/a11y_breadcrumbs

But fascinating the HTML spec has it in paragraphs. I'm not 100% opposed to the idea, but I think the ol does provide a bit of a sense of hierarchy (like the first item is the 1st level page and the 2nd item is 2nd level)

nathan-barrett:
Oh, nice - thanks for the links @craigcook - I'll take a look and see which direction makes the most sense. Maybe an `ol` rather than a `ul` at the very least.

src/patterns/molecules/breadcrumb/breadcrumb.hbs
nathan-barrett: Yeah I think you're right.

craigcook:
Yeah, an ordered list feels right to me. Let's stick with that.

maureenlholland (approved):
Pending rebase and approval from @craigcook, but everything related to breadcrumb looks good to me! r+ 🏆

craigcook (approved):
LGTM! 🍞

PR #734, opened 2021-10-14
Remove sticky nav for reduced motion (fixes #733)

### LINKED ISSUE #733: Way too much pain/animation on Mozilla websites
## Description

When I visit Mozilla support websites, I get hit with a ton of pain/animation, and get migraines, dizziness, and nausea from these.

### Steps to reproduce

visit https://www.mozilla.org/en-US/

scroll down

The navigation header suddenly scrolls up faster than the page.

scroll down at least a quarter of the way, scroll up, scroll down again.

The navigation header scrolls down and can bob up and down, giving me migraines. This appears to use Javascript, so css fixes aren't possible.

### Expected result

At this point, I expect websites to include a lot of extra pain/animation. But it would be really helpful to cut down on the pain/animation, avoid sticky elements, and/or check for prefers-reduced-motion.

### Actual result

The navigation header suddenly scrolls up faster than the page.

The navigation header scrolls down and can bob up and down, giving me migraines. This appears to use Javascript, so css fixes aren't possible.

### Environment

What browser and operating system are you using?

Firefox 93 in MacOS 10.14.6

P.S. I can't scroll the documentation here, because it has a non-scrolling sidebar: https://protocol.mozilla.org/patterns/organisms/navigation.html

maureenlholland:
## Description

We want to remove sticky behaviour from the nav for users who prefer reduced motion.

### Issue
https://github.com/mozilla/protocol/issues/733

### Testing

You'll have to toggle your preference for reduced motion according to your user agent: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion#user_preferences

http://localhost:3000/demos/navigation.html

When you prefer reduced motion, the nav should be static.
When you do not prefer reduced motion, the nav should be sticky with animations to scroll in and out of view on page scroll (current behaviour all the time on https://www.mozilla.org/en-US/).

ℹ️ I have a fix for this specifically in bedrock: https://github.com/mozilla/bedrock/pull/10592
Do we want to merge that fix for now and save this until we can fully backport the bedrock nav as well?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ## Bug Fixes
 
+* **js** when navigation has `mzp-is-sticky` class, respect user preference for reduced motion (#733)
 * **css:** Replace deprecated `/` division operator with `math.div()` (#722)
 * **css:** Set Split media width to 100% to accommodate responsive video (#711)
 
--- a/src/assets/js/protocol/protocol-navigation.js
+++ b/src/assets/js/protocol/protocol-navigation.js
@@ -213,13 +213,20 @@ if (typeof window.Mzp === 'undefined') { // eslint-disable-line block-scoped-var
          * If there are multiple navigation organisms on a single page,
          * assume only the first (and hence top-most) instance can and
          * will be sticky.
+         *
+         * Do not init sticky navigation if user prefers reduced motion
          */
-        _navElem = document.querySelector('.mzp-c-navigation');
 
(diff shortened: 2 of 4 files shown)


craigcook (approved):
🙈

craigcook:
The nav was redesigned some time ago but only in bedrock and never backported because we didn't have an adequate replacement for the sub-links, and now it's being redesigned again so that nav will never be backported. I think we can land this in Protocol now and also make the same updates in bedrock, since that nav is a standalone component separate from Protocol.

PR #737, opened 2021-10-25
Correct Split docs, body width class goes on the container

craigcook:
## Description

Noticed a mistake in the Split notes. The body width class goes on the outer container, not the body container (because grid). Just a doc fix, doesn't need to go in the changelog.

### Testing


craigcook:
This is just a doc fix so I feel OK skipping review and merging it myself, even though the button is red and scary.

PR #741, opened 2021-11-04
Add tertiary theme color, rename alt to secondary

craigcook:
## Description

Adds a tertiary "accent" background color to the Firefox and Mozilla themes. I'm not sure if "accent" is actually the best name for this, happy to discuss alternatives.

### Testing
https://demo1--mozilla-protocol.netlify.app/demos/theme-firefox.html
https://demo1--mozilla-protocol.netlify.app/demos/theme-mozilla.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,15 @@
+# HEAD
+
 ## Features
 
+* **css:** Add tertiary theme colors.
+* **css:** (breaking) Rename "alt" theme colors to "secondary."
 * **component:** New breadcrumb component.
+
+## Migration Tips
+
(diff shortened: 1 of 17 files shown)


craigcook:
I think we're going to rename the "alt" colors to "secondary" and rename "accent" to "tertiary"

craigcook:
Updated to rename "alt" to "secondary" which makes this a breaking change. We could alias the "alt" name for backwards compatibility and avoid a breaking change, but then we'd need to keep those aliases indefinitely so we might as well just get it over with now.

nathan-barrett (approved):
everything looks good to me now, r+ 🧇

PR #743, opened 2021-11-09
Fix image overlapping content in a reversed Split with media overflow

craigcook:
## Description

A Split component with both reversed layout *and* overflowing media allows the media to overlap the content. The layout is reversed but when the image has no max-width, it's naturally left-aligned and expands to the right, often covering some of the text.

Floating the image to the right in this scenario aligns it to the right and lets it overflow to the left (or flippity-floppity in RTL).

### Testing
See it in action on http://localhost:3000/demos/split-visual.html in the fifth example down.


maureenlholland (approved):
r+ 🐬

PR #750, opened 2021-12-07
Add more robust polyfill (fixes #736)

### LINKED ISSUE #736: Type error: e.matches is not a function in protocol-utils.js
## Description

One of the highest-frequency JS errors we currently see on bedrock is from protocol-utils.js (57k + errors):

https://sentry.prod.mozaws.net/operations/bedrock-prod/issues/9874490/

It seems to be this line: https://github.com/mozilla/protocol/blob/main/src/assets/js/protocol/protocol-utils.js#L40

### Steps to reproduce

The error occurs in older browsers that do not support element.matches: https://developer.mozilla.org/en-US/docs/Web/API/Element/matches

### Expected result

Could this function fail more gracefully if the browser does not support this function?

Perhaps the [details component](https://github.com/mozilla/protocol/blob/main/src/assets/js/protocol/protocol-details.js) that uses the `nextUntil` function could add a feature detect to return early if it fails?

### Actual result

The script throws a JS error and halts execution.

### Environment

Browser support can be seen here: https://developer.mozilla.org/en-US/docs/Web/API/Element/matches

maureenlholland:
## Description

Adds a more robust polyfill (particularly `mozMatchesSelector`) to catch more e.matches errors and avoid overly noisy Sentry reporting.

### Issue

https://github.com/mozilla/protocol/issues/736

### Testing

https://sentry.prod.mozaws.net/operations/bedrock-prod/issues/9874490/tags/

Test on browserstack
Win 7 + Firefox 30 (most frequent tags in this error)
Win 10 + IE 8

Live site should show `.matches` error in console on an older browser: https://protocol.mozilla.org/patterns/molecules/details.html
live

Local site should show no error with same browser: http://localhost:3000/patterns/molecules/details.html
local

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# HEAD
+
+## Bug Fixes
+
+* **js:** use more robust polyfill for `e.matches` (#736)
+
 # 16.0.0
 
 ## Features
--- a/src/assets/js/protocol/protocol-utils.js
+++ b/src/assets/js/protocol/protocol-utils.js
@@ -13,9 +13,24 @@ if (typeof window.Mzp === 'undefined') { // eslint-disable-line block-scoped-var
 
     var Utils = {};
 
-    // matches() vendorfill, used by nextUntil
+    /**
+     * Element.matches() polyfill (IE8 support)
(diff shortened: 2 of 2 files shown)


craigcook (approved):
🐻‍❄️

PR #757, opened 2021-12-15
Fix keyboard focus capture on modal open animation (Fixes #749)

### LINKED ISSUE #749: Modal keyboard focus logic is broken
## Description

When doing some keyboard navigation testing, I noticed the modals on bedrock no longer limit focus to elements within the modal content.

### Steps to reproduce

1. Open https://www.mozilla.org/en-US/about/leadership/
2. Click on a leadership photo to open a modal.
3. Press the tab key to start moving the focused element.

### Expected result

Only elements within the modal should be focusable. The modals should behave like they do in the Protocol demo: https://protocol.mozilla.org/patterns/organisms/modal.html

### Actual result

It is possible for any element in the web page to gain focus.

### Environment

Firefox Nightly

alexgibson:
## Description

Adds a small delay before calling `modal.focus()` when opening the Modal component.

- ~[ ] I have documented this change in the design system~.

### Issue

#749

### Testing

The bug is hard to reproduce on the Protocol demo page below, because I'm assuming it's a less intensive page to animate? You can reproduce the bug in the www.mozilla.org links in the attached issue.

I tested locally in bedrock that this fix works (feel free to do the same in review).

http://localhost:3000/patterns/organisms/modal.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,13 @@
 # HEAD
 
+## Bug Fixes
+* **js:** Fix keyboard focus capture on modal open animation (#749).
+
 ## Features
 
-* **node:**  Create a standalone Webpack config for compiling the Protocol NPM package (#746).
-* **js:**  Update protocol to extend standard ESLint configs (#753)
-* **js:**  Move Karma test command to NPM script (#748).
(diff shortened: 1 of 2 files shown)


craigcook (approved):
⏲️

Issue #760, opened 2022-01-05
Use WebPack for tasks


craigcook:
Protocol currently uses Gulp as a task runner but we're taking the Fractal migration as an opportunity to switch to WebPack.

An off-top-of-the-head list of tasks:

- Compile Protocol SCSS
- Compile Fractal theme SCSS
- Build for local dev: 
  - Lint, compile, watch Protocol SCSS
  - Lint, watch Protocol JS
  - Lint, compile, watch Fractal theme SCSS
  - Copy Protocol assets to static
  - Fractal start
- Build for prod:
  - Lint, compile Protocol SCSS
  - Lint JS
  - Lint, compile Fractal theme SCSS
  - Copy Protocol assets to static
  - Fractal build
- Build for publication to NPM
  - Lint, compile Protocol SCSS
  - Lint JS
  - Lint SCSS
  - Lint JS
- Copy Protocol assets to static

alexgibson:
I think we can cut down / simplify this list based on some of the work already done recently:

- https://github.com/mozilla/protocol/issues/746. This created a webpack config for building the NPM package. This exists entirely standalone from Fractal (and imho should probably remian so).
- https://github.com/mozilla/protocol/issues/745. This removed existing linting tasks from Gulp and replaced them with NPM commands. This means Webpack / Fractal no longer need to factor in linting, as it can be run as an NPM [pre-script](https://docs.npmjs.com/cli/v6/using-npm/scripts#pre--post-scripts) before any other script.
- https://github.com/mozilla/protocol/issues/748. This did the same for JS tests as we did for linting, removing the need for Gulp/Webpack to run them directly.

Based on the above, I think we can simplify this list to:

- Build for local dev:
  - Copy Protocol JS / assets to static.
  - Compile Protocol SCSS.
  - Compile Fractal theme SCSS.
  - Watch SCSS/ JS assets for changes.
  - Fractal start
- Build for prod:
  - Copy Protocol JS / assets to static.
  - Compile Protocol SCSS.
  - Compile Fractal theme SCSS.
  - Fractal build

It looks like [Fractal](https://fractal.build/guide/getting-started.html) comes with its own dev server and watch command, so the code for both local-dev and prod commands can likely be shared.

If we still need Webpack to compile CSS for the docs and copy JS/static assets, then I'd suggest we use a new config file such as `webpack.docs.config.js` to keep things separate from the NPM package stuff.

PR #761, opened 2022-01-18
Add webpack with tasks for build and static

craigcook:
## Description
This adds webpack as a dependency and also migrates the tasks for linting CSS and JS files, building the package, and copying static files.

### Issue
#760 

### Testing
Do the webpack static and build: `npm run webpack`
Lint CSS: `npm run lint-css`
Lint JS: `npm run lint-js`
Lint both: `npm run lint`


craigcook:
So instead of treating this as a separate PR I was able to rebase the `fractal` branch from main and do the necessary post-rebase fixes there. But this mean the `fractal` branch could be wonky if I messed something up in the rebase, but we can always just fix it.

Issue #763, opened 2022-02-05
URL to heading elements in docs routes to localhost


kimberlythegeek:
## Description

On [Section Heading](https://protocol.mozilla.org/patterns/atoms/section-heading.html) page in the docs, the "heading element" URL routes to localhost:3000

PR #764, opened 2022-02-05
fix url for heading element

kimberlythegeek:
## Description

Fix url in https://protocol.mozilla.org/patterns/atoms/section-heading.html that was routing to localhost

### Issue

#763


craigcook:
This was fixed in https://github.com/mozilla/protocol/commit/fb1a92935a954e422f357398562eac99038ed835 but evidently the fix hasn't been deployed to production, and prod is actually a several commits behind main. We'll get that sorted and update the live site.

Issue #774, opened 2022-03-08
Add mzp-c-card-large class to hero card layouts


reemhamz:
### Please describe the use case(s) for this component
When using the card component's layout styles, specifically when using the`mzp-l-card-hero` layout, it's not explicitly said that we should add `mzp-c-card-large` class to the card we want to be enlarged. It would create less friction if documented.

avi5079:
Hello @reemhamz,
I am a beginner to open source. I would like to learn by contributing to this repo. Could you please guide me with this issue like how to get started?

reemhamz:
Hi @avi5079, nice to see an interest in contributing to our repository!
The official website for this repository is this: https://protocol.mozilla.org/
Follow the installation instructions for Protocol, as seen in the readMe here: https://github.com/mozilla/protocol

You'll need to create your own branch in order for us to review.

I was thinking of adding a sentence in this page for the [card layout](https://protocol.mozilla.org/patterns/templates/card-layout.html) templating detailing that we should add `mzp-c-card-large` class to the larger sized image in the `mzp-l-card-hero` class.

You'll need to understand how the different card layouts and classes work with our Protocol design system. I would suggest playing around in devtools and applying different classes to the [card layout demo page](https://protocol.mozilla.org/demos/card-layout.html). We use Protocol heavily in our mozilla.org repository, [Bedrock](https://github.com/mozilla/bedrock). So you can even locally install that repository and play around with how we have Protocol implemented throughout the site.

To better understand how we use Protocol's design system in Bedrock, here's the [documentation detailing it](https://bedrock.readthedocs.io/en/latest/coding.html#working-with-protocol).

Let me know if you have any further questions.

avi5079:
Just to confirm, after learning about different card layouts, I have to edit the https://protocol.mozilla.org/patterns/templates/card-layout.html page specifying the detail you mentioned above? . Should I insert it under the Tips subheading?

avi5079:
Thank You for the clarification, @reemhamz.

avi5079:
Hey @reemhamz 
Is there any other issue/feature I can work on?

PR #776, opened 2022-03-30
Added documentation for card layout templating

avi5079:
## Description

Added documentation detailing that we should add mzp-c-card-large class to the larger sized image in the mzp-l-card-hero class in the card layout template page.

### Issue

#774 

### Testing

mozilla


avi5079:
review @reemhamz

PR #781, opened 2022-05-05
Add form components

craigcook:
## Description
Adds form component examples and documentation, plus a few small theme tweaks.

https://demo1--mozilla-protocol.netlify.app/


reemhamz:
Good job with this! :) 
I like the addition of the error messages, looks super neat.
One nit and one suggestion:

**Nit:** I see you used `Fieldset` twice and created two different sections with similar information, but different examples of presenting it. Was this intentional?

**Suggestion:** I was thinking of changing the order of where you placed the `Titles` section, I would have it moved up a couple points because having it at the bottom of the list could be forgotten. Probably can be placed around the first `Fieldset` section?

craigcook:
> **Nit:** I see you used `Fieldset` twice and created two different sections with similar information, but different examples of presenting it. Was this intentional?

One is basic styling of the `fieldset` element (really just `legend`) and the other is a component with a few more supporting classes, we just called the component a Field Set as well. We could clarify the difference in the docs, but now that I think about it I'm not sure we need to show an example of the basic `fieldset` styles at all. Maybe I'll just remove that.

> **Suggestion:** I was thinking of changing the order of where you placed the Titles section, I would have it moved up a couple points because having it at the bottom of the list could be forgotten. Probably can be placed around the first Fieldset section?

Agreed, I'll move titles up.

Issue #759, opened 2022-01-05
Migrate Protocol documentation to Fractal


craigcook:
The Protocol doc site (https://protocol.mozilla.org) was initially built with Drizzle, a platform that has since been officially deprecated and will no longer be updated. We've opted to migrate to Fractal (https://fractal.build) for our documentation but that migration will entail a number of steps.

This issue can serve as an epic tracking the full migration and we'll break it down into individual tasks under this umbrella.

Project board: https://github.com/mozilla/protocol/projects/10

craigcook:
We have a branch for all the Fractal migration work: https://github.com/mozilla/protocol/tree/fractal.

Though it started as a branch from `main`, the first thing I did was delete almost everything except the core assets and then started adding new things from scratch. So the new branch is such a radical diversion that it hardly seems worth trying to merge it to main. Instead, my idea is to eventually archive the `main` branch and rename the `fractal` branch to `main`. 

So meanwhile we still have a canonical `main` branch that reflects the current state of Protocol and we can do all our migration work in the `fractal` branch. We should avoid doing much real work in `main` but could still add any important fixes there that need to be published before Fractal is ready (and we may need to migrate those fixes to the `fractal` branch by hand).

I've been making direct commits to `fractal` so far but now that a sort of baseline has been established I intend to start making pull requests to that branch for review.

reemhamz:
Hey developers, here's how we run this branch:
First, make sure you run `npm install`. You're going to need 2 terminal windows running.
- In one, type `npm run webpack-docs`
- In the other, type `npm start` and follow the localhost URL.

craigcook:
Let's edit the checklist comment above to put our names next to any components we're claiming, to avoid more than one person working on the same thing.

We could also create an issue for each one but that feels like a lot.

reemhamz:
Very odd, I didn't get that either.

nathan-barrett:
I actually got this, but it happened when I ran `npm build` on the `main` branch

reemhamz:
Are we supposed to run `npm build`?

craigcook:
> Are we supposed to run `npm build`?

No, just `npm start`. The fractal branch doesn't have a `src` folder at all, so if anything that's a leftover from switching branches. We could add it to gitignore if this keeps happening.

nathan-barrett:
Sorry, no - I was just seeing how the webpack build worked in protocol to help when I was working on trafficcop

maureenlholland:
[...]

PR #786, opened 2022-06-02
Some components migration into Fractal

reemhamz:
## Description
Migrating some components into Fractal. I'm unsure of how many I'll be doing in this one PR, but I'll mark them as I go below 👇🏽

- [ ] Breadcrumb
- [ ] Language switcher
- [ ] Menu list
- [ ] Article
- [ ] Content container

### Issue

https://github.com/mozilla/protocol/issues/759

### Testing

You're going to need 2 terminal windows running.

- In one, type `npm run webpack-docs`
- In the other, type `npm start` and follow the localhost URL.

#### URLs to test:
- [ ] http://localhost:3000/components/detail/breadcrumb
- [ ] http://localhost:3000/components/detail/language-switcher
- [ ] http://localhost:3000/components/detail/menu-list--default
- [ ] http://localhost:3000/components/detail/menu-list--download
- [ ] http://localhost:3000/components/detail/article
- [ ] http://localhost:3000/components/detail/content-container


components/breadcrumb/breadcrumb.html
```diff
@@ -0,0 +1,10 @@
+<nav aria-label="breadcrumb" class="mzp-c-breadcrumb">
+  <ol class="mzp-c-breadcrumb-list">
+    <li class="mzp-c-breadcrumb-item">
+      <a href="#">Parent page</a>
```
reemhamz, line 4: Since the content in this page is very... minimal, I opted to just hard-code it here and make a `readme.md` for the page notes instead of creating a `breadcrumbs.config.yml` file.

components/language-switcher/language-switcher.html
```diff
@@ -0,0 +1,33 @@
...
+    <option value="fr">Français</option>
+    <option value="es-ES">Español</option>
+    <option value="ja">日本語</option>
+  </select>
+  <!-- form submit button is shown when JavaScript is not enabled -->
+  <button type="submit">Go</button>
+</form>
+
+<script
+ src="../../../assets/protocol/protocol/js/protocol-lang-switcher.js">
```
craigcook, line 19: This path should be:
```suggestion
 src="{{ '/protocol/js/protocol-lang-switcher.js' | path }}">
```

components/language-switcher/language-switcher.html
```diff
@@ -0,0 +1,33 @@
+<form class="mzp-c-language-switcher " method="get"
+ action="#">
```
craigcook, line 2: There are some awkward line breaks in the markup, I assume copied from how the code is formatted on https://protocol.mozilla.org/patterns/molecules/language-switcher.html. But we can tolerate longer lines than whatever that formatter is set for, so this could be tidied up a bit. You might prefer to copy from the original sources (https://github.com/mozilla/protocol/blob/main/src/patterns/molecules/language-switcher/language-switcher.hbs) and convert the Handlebars to Nunjucks.

craigcook (changes requested):

craigcook:
I added `protocol-base.js` to the preview template. If you rebase from the fractal branch it should now add the `js` class to hide your go button.

components/menus/menu-item/menu-item.html
```diff
@@ -0,0 +1,38 @@
...
+# url           - string
+# class         - string
+# required      - boolean
+# disabled      - boolean
+# selected      - boolean
+#}
+<section class="mzp-c-menu-item {% if icon %}mzp-has-icon{% endif %} {% if class %}{{ class }}{% endif %}">
+  {% if url %}<a class="mzp-c-menu-item-link" href="{{url}}">{% else %}<div class="mzp-c-menu-item-head">{% endif %}
+    {% if icon %}
+      {% block icon %}
```
craigcook, line 17: Although Nunjucks supports blocks (same as Jinja), I'm not sure Fractal supports them in components, or at least it didn't work when I tried it (I might have done something wrong). So this might be a shortcoming we'll have to work around. Just a heads-up.

craigcook: I was wrong! Fractal does support blocks, you just have to be sure you tell the child template what parent template it's extending, e.g. `{% extends '@menu-item' %}` at the top. But this doesn't seem compatible with Fractal's `render` helper so it may not be very useful to us in the end.

craigcook:
Menu and Menu-Item are actually part of the global navigation, and Menu-List is a separate component, so "Menus" isn't really the right category to group these under. 

I'd recommend pulling Menu and Menu Item out of this PR since they probably need to be migrated along with the rest of the navigation stuff (and it's possibly blocked on getting cards migrated first anyway).

components/language-switcher/readme.md
```diff
@@ -0,0 +1,8 @@
+A UI molecule for changing the language of the current page.
```
craigcook, line 1: We should phase out the atomic design jargon and just call everything a "component."

components/menus/menu-list/menu-list.html
```diff
@@ -0,0 +1,15 @@
+<div class="mzp-c-menu-list {% if class %}{{ class }}{% endif %}">
+  <h3 class="mzp-c-menu-list-title">Meet the Family</h3>
```
craigcook, line 2: Maybe we should use more future-proof and generic examples. Lockwise and Send are no longer products so this looks a little dated.

components/menus/menu-list/readme.md
```diff
@@ -0,0 +1 @@
+A sort title, when clicked, expands a list of links.
```
craigcook, line 1: It would be nice to flesh out this documentation with some more description of how it works, suggestions on uses, mention how to use the JS to initialize it, etc. 

Not a blocker for merging since we can always add more later, but this is pretty scant documentation (I know it was just taken from the original; that one should have been better documented too).

craigcook (changes requested):

components/language-switcher/readme.md
reemhamz: Oh, are we completely ditching the atomic design lingo? Guessing all the components will be under one header then?

craigcook: Yeah... atomic lingo is very useful for discussing design systems but I'm not sure it's useful as an information architecture strategy.

components/content-container/readme.md
```diff
@@ -0,0 +1,15 @@
+This container element sets the maximum width of content on the page, centers the column horizontally in the viewport, and includes some padding that prevents its contents from butting against the edges of the viewport or against other adjacent elements. The maximum width and spacing change at different responsive breakpoints to adapt to different viewport widths.
+
+- The container's width isn't explicit; it's a fluid box with a minimum width of `304px` and constrained to a maximum width that differs at different responsive breakpoints. If you want to know how wide this box is, the answer is: it depends.
+- Side padding also adjusts at different breakpoints, with less padding in small windows and more padding in large windows.
+- You can apply theme classes to set a max-width narrower than the default, to constrain the maximum width even in larger viewports.
+    - `mzp-t-content-sm` : `432px`
+    - `mzp-t-content-md` : `688px`
+    - `mzp-t-content-lg` : `928px`
+    - `mzp-t-content-xl` : `1152px`
```
reemhamz, line 9: Should I create variations for each of these classes? It would be neat to showcase all the different viewports!

reemhamz: Update: added variants in my latest commit.

components/content-container/readme.md
```diff
@@ -0,0 +1,15 @@
+This container element sets the maximum width of content on the page, centers the column horizontally in the viewport, and includes some padding that prevents its contents from butting against the edges of the viewport or against other adjacent elements. The maximum width and spacing change at different responsive breakpoints to adapt to different viewport widths.
+
+- The container's width isn't explicit; it's a fluid box with a minimum width of `304px` and constrained to a maximum width that differs at different responsive breakpoints. If you want to know how wide this box is, the answer is: it depends.
+- Side padding also adjusts at different breakpoints, with less padding in small windows and more padding in large windows.
+- You can apply theme classes to set a max-width narrower than the default, to constrain the maximum width even in larger viewports.
+    - `mzp-t-content-sm` : `432px`
```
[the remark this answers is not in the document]
craigcook: Maybe... it's not always code, but sometimes it is (like in this case it's technically a CSS value). I don't know that we need a rule for it but it doesn't hurt either.

components/content-container/content-container.config.yml
```diff
@@ -0,0 +1,24 @@
+context:
+  label: Content container
+default: default
+variants:
+  - name: Small container
+    notes: |
+      A small container of `mzp-t-content-sm` class with `max-width: 432px`
```
craigcook, line 7: Grammar nit: 
```suggestion
      A small container, adding the class `mzp-t-content-sm` to set `max-width: 432px`
```

craigcook: Notes on the other size variants should also be updated.

components/language-switcher/language-switcher.html
```diff
@@ -0,0 +1,35 @@
+{# @select
```
craigcook, line 1: This should be the component handle, `@language-switcher`

components/language-switcher/language-switcher.html
```diff
@@ -0,0 +1,35 @@
...
+# parameters:
+# id            - string, required
+# name          - string, required
+# label         - string, required
+# class         - string
+#}
+<form class="mzp-c-language-switcher {% if class %}{{ class }}{% endif %}" method="get" action="#">
+  <a class="mzp-c-language-switcher-link" href="https://www.mozilla.org/locales/" target="_blank" external="noopener noreferrer">Language</a>
+  <label for="mzp-c-language-switcher-select">Language</label>
+  <select id="mzp-c-language-switcher-select" class="mzp-js-language-switcher-select" name="{{ name }}">
```
craigcook, line 12: I think this could be an instance of the regular form select, and can even reuse the same context data:

`{% render '@select', { id: 'mzp-c-language-switcher-select', class: 'mzp-js-language-switcher-select', name: 'language', label: 'Language', options: '@select.options' } %}`

If you do that I don't think this component needs a config file at all. Also the parameters above would be reduced just to `class`. Though thinking about it, I'm not sure this component needs an additional class. It doesn't have any variants so any additional classes would be some kind of custom thing. So maybe better to remove that for simplicity.

components/language-switcher/readme.md
```diff
@@ -0,0 +1,8 @@
+A component for changing the language of the current page.
+
+### Notes
+- Assumes the first path in the page URL immediately after hostname is a valid locale code e.g. https://www.mozilla.org/en-US/firefox/.
+- This molecule can be displayed inside the Footer organism, but could also be used elsewhere if more appropriate.
```
craigcook, line 5: Let's rewrite without the atomic design lingo.
```suggestion
- This component can be displayed inside the Footer component, but could also be used elsewhere if more appropriate.
```

craigcook (changes requested):

craigcook:
Don't add any more components to this PR so we can get this batch wrapped up and merged. Nice work!

components/language-switcher/readme.md
reemhamz: d'oh! good catch.

components/content-container/readme.md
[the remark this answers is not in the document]
reemhamz: I'm happy to hear what the other front-enders think? That way we can make sure it's consistent throughout the new documentation

craigcook (approved):

components/content-container/readme.md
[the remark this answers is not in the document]
nathan-barrett: I think that I tend to prefer use the `` for pixel values. Even if its not code - the styling of it helps my eye gravitate towards it. But I'm open to either one - as long as its consistent? I wouldn't mind @maureenlholland idea of a context specific rule.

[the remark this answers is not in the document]
maureenlholland: The context specific rule is a minor preference at a push and might be difficult to keep consistent if we always have to explain the rule. Looks like there's a general preference for pixels in code syntax and I'm happy to see that become a general rule for the sake of clarity and consistency.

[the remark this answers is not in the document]
reemhamz: Thanks for your input team :) let's test out wrapping pixels in code syntax for now. I'm with @nathan-barrett here, where it helps my eyes gravitate a lot better. I can create a PR for this sometime tomorrow.

PR #789, opened 2022-06-07
Migrate Picto Card

craigcook:
## Description
Migrates the Picto Card component (which is deprecated, but still in Protocol so still needs to be documented).

### Testing
http://localhost:3000/components/detail/picto-card--default
http://localhost:3000/components/detail/picto-card--compact


maureenlholland (approved):
r+ 🔴

PR #790, opened 2022-06-08
Migrate Details and Details component

craigcook:
## Description
Migrates the (native) details element and Details component.

### Testing
http://localhost:3000/components/detail/details
http://localhost:3000/components/detail/details-component--default

### DIFF
--- a/components/details-component/details-component--custom.readme.md
+++ b/components/details-component/details-component--custom.readme.md
@@ -1,0 +1,7 @@
+The functions driving the Details component can be called independently for any container
+element. It won’t recieve any styling without the Protocol classes so you’ll have to
+style it separately (styles in this demonstration are intentionally tacky).
+
+Initialized using `Mzp.Details.init(selector, options)`, where `options` is an Object with the following properties:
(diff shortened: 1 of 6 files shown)


components/details-component/details-component--custom.readme.md
```diff
@@ -0,0 +1,7 @@
+The functions driving the Details component can be called independently for any container
+element. It won’t recieve any styling without the Protocol classes so you’ll have to
```
maureenlholland, line 2:
```suggestion
element. It won’t receive any styling without the Protocol classes so you’ll have to
```

maureenlholland (changes requested):
Typo fixes. One suggestion.

Non-blocking but worth discussion: Organization and naming-wise, this is tricky. Is there a way in config to specify a sidenav name different than the file name? "Components" > "Details Component" looks a bit awkward since nothing else in the list needs that "Component" distinction. Non-native Details? Custom Details? Flexible Details?

maureenlholland (approved):
r+ 🪗

PR #791, opened 2022-06-08
Migrate sidebar menu and main with sidebar

maureenlholland:
## Description

Migrates sidebar menu component and main with sidebar template

## Issue
https://github.com/mozilla/protocol/issues/759

### Testing

http://localhost:3000/components/detail/main-with-sidebar--default
http://localhost:3000/components/detail/main-with-sidebar--sidebar-right
http://localhost:3000/components/detail/example-sidebar-menu

http://localhost:3000/components/detail/sidebar-menu

### DIFF
--- a/components/main-with-sidebar/example-sidebar-menu/example-sidebar-menu.html
+++ b/components/main-with-sidebar/example-sidebar-menu/example-sidebar-menu.html
@@ -5,8 +5,40 @@
     {% render '@sidebar-menu' %}
   </aside>
   <main class="mzp-l-main">
-    <p>Main content.</p>
+    <h1>Welcome to the main content</h1>
+
     <p>The sidebar comes <strong>before</strong> the main section in the source order.</p>
+
+    <h2>Heading Level 2</h2>
+
+    <p>Et pri porro ceteros tacimates, est at option dignissim intellegat.
+      Tollit fuisset urbanitas nec no, feugiat sensibus an mea, nominavi
+      <a href="#">interpretaris te eam</a>. No virtute imperdiet pro, ut pro
+      sint viderer philosophia. Torquatos necessitatibus sit in, ad sit summo
+      molestie adolescens, liber oratio veniam in per.</p>
+
+    <p>No qui option timeam impetus, dolor salutatus pertinacia ea eam, at
+      inani summo democritum vis. Est id quis admodum tractatos, ei ius
+      eligendi platonem, maiorum eleifend et eam. Ei odio cetero voluptatibus
+      ius, prima blandit epicurei te cum. Illud verterem principes vix te,
+      te pro tale dicit. Debitis senserit no quo, vix et altera mollis
+      scaevola. Sit no soleat aperiri vituperata. Qui <a href="#">id placerat
+      euripidis</a>.</p>
+
+    <h3>Heading Level 3</h3>
+
+    <p>An quo epicurei placerat repudiare. Ut cetero molestiae mei, per id
+      efficiantur consectetuer, in paulo repudiare vim. Inermis delectus
+      adipiscing eos ad, sed ei graeci eleifend intellegam. Eu abhorreant
+      neglegentur duo, eu mei nisl mundi.</p>
+
+    <p>Suavitate comprehensam ea nam, tacimates vituperata vim te. Usu in
+      laboramus maiestatis definitionem, noster complectitur at vim. Qui ut
+      doctus sententiae percipitur, vivendum mediocritatem ea eum. Eu cum
+      tantas nostrud, ea docendi blandit nec, vis id aeterno intellegat
+      accommodare.</p>
   </main>
 {% endblock %}
+
+{% block styles %}{% endblock %}
 
--- a/components/main-with-sidebar/01-main-with-sidebar/main-with-sidebar.html
+++ b/components/main-with-sidebar/01-main-with-sidebar/main-with-sidebar.html
@@ -11,6 +11,7 @@
   {% endblock %}
 </div>
 
+{% block styles %}
 <!-- Inline styles are for demo purposes only. Prefer external stylesheet in production. -->
 <style>
   .mzp-l-main {
@@ -21,4 +22,5 @@
     border: 4px solid lightseagreen;
   }
 </style>
+{% endblock %}
 


components/main-with-sidebar/main-with-sidebar.html
```diff
@@ -0,0 +1,33 @@
...
+    <p>Main content.</p>
+    <p>The sidebar comes <strong>after</strong> the main section in the source order.</p>
+  </main>
+
+  <aside class="mzp-l-sidebar">
+    <p>Sidebar content.</p>
+  </aside>
+</div>
+
+<!-- Inline styles are for demo purposes only. Prefer external stylesheet in production. -->
```
maureenlholland, line 23: Completely arbitrary colors here. Also, not sure what to do with variants. It requires a right or left layout, so neither is really "default" and it felt important to display them both, but that kind of breaks the render handle functionality, so maybe not worth it? Or move Right Sidebar to a variant? Very open to recommendations here.

components/sidebar-menu/readme.md
```diff
@@ -0,0 +1,5 @@
+Local navigation for a subset of pages or section of a website, intended to appear in a sidebar adjacent to the main content. It can accommodate multiple categories/sub-sections in a two-level hierarchy with titles corresponding to the top level page of each sub-section.
+
+The menu summary reflects the current page, shown when the menu is collapsed in small viewports. The menu label should be the title of the site section.
+
+- Indicate the current page with the `mzp-is-current` class.
```
reemhamz, line 5: **Nit:** put `Tips` on top as a header for this point

components/main-with-sidebar/main-with-sidebar.html
```diff
@@ -0,0 +1,33 @@
...
+
+  <aside class="mzp-l-sidebar">
+    <p>Sidebar content.</p>
+  </aside>
+</div>
+
+<div class="mzp-l-content mzp-has-sidebar mzp-l-sidebar-right">
+  <main class="mzp-l-main">
+    <p>Main content.</p>
+    <p>The sidebar comes <strong>after</strong> the main section in the source order.</p>
```
reemhamz, line 15: **Suggestion:** I know you moved the last point from this file's notes to be reflected in this view (good idea btw, I got a little lost fiddling with sidebar when I used it for something), however I think it would also be good to keep it in the notes section, so it can stay as a one-source-of-truth for this layout?

components/main-with-sidebar/main-with-sidebar--sidebar-menu.html
```diff
@@ -0,0 +1,9 @@
+<div class="mzp-l-content mzp-has-sidebar mzp-l-sidebar-left">
```
reemhamz, line 1: **Praise:** I like that you created this file to show a realistic demo. I'm happy to keep it (it's super useful imo), however I do know that it's not necessarily a variant, which is what we've been using these page views for. @craigcook thoughts?

components/main-with-sidebar/readme.md
```diff
@@ -0,0 +1,15 @@
+A block with a main column and one sidebar on either the left or right. A single page should only have one main section and one sidebar.
+
+The columns will stack in small viewports, form a 1/3-2/3 split in medium sized viewports, and a 1/4-3/4 split in wider viewports (limiting the width of the sidebar).
+
+This layout template only sets up the basic columns. The main content column is quite wide in large viewports and can result in uncomfortably long lines of text. You may want to use other nested elements to limit line lengths.
+
+- Layout class `mzp-l-sidebar-left` or `mzp-l-sidebar-right` is required.
```
reemhamz, line 7: **Nit:** place a `Tips` header on top of this line

components/main-with-sidebar/main-with-sidebar.html
reemhamz: Definitely important to display them both! However, going with how Craig setup the picto variants (http://localhost:3000/components/detail/picto--default), each type of picto/side of picto has its own variant to be shown. 
It would have probably been better to have the LTR/RTL options like we do in Drizzle, but we don't have that set up just yet. So I would say just stick to the variants to make it more consistent with how other components are laid out.

reemhamz: Also, TIL the CSS color `papayawhip`. Nice 🍦

reemhamz (changes requested):
Just a few nits and a couple things worth discussing. Good job putting this together otherwise! :papayawhip:

components/main-with-sidebar/main-with-sidebar.html
maureenlholland: > It would have probably been better to have the LTR/RTL options like we do in Drizzle

This wouldn't entirely do what we need here, which is show that there are two layout classes available. The LTR or RTL toggle would only show how those classes respect the language direction. (i.e. `mzp-l-sidebar-right` would move to the the left side on a RTL direction)

re: consistency with variant use 💯 - will update

components/main-with-sidebar/main-with-sidebar.html
maureenlholland: It's one line and this isn't a true variant, so I'm easily swayed here, but my general feeling is it's better to break up documentation where possible and not adhere to a comprehensive one-source-of-truth on the default notes view. It's a feeling based on personal preference, as I know my eyes glaze over a bit when docs go too long.

components/main-with-sidebar/main-with-sidebar--sidebar-menu.html
craigcook: Given that we don't currently have a replacement for the old demo pages, I think we can still include useful examples in Fractal, but I agree they probably don't count as "variants". 

But we can add a template with a different filename (not a `--` variant) in the same folder and Fractal will treat it as another "component" in the UI rather than a variant. I did that with an example form: https://demo1--mozilla-protocol.netlify.app/components/detail/example-form.html

components/main-with-sidebar/main-with-sidebar.html
reemhamz: Ah, yes that's a reminder I needed.

reemhamz (approved):
Cool!
Just a suggestion I left that I think is worth changing. Everything else looks wonderful. Good job! 🏄🏻 r+wc

components/main-with-sidebar/main-with-sidebar.html
```diff
@@ -1,22 +1,24 @@
...
+      <p>Sidebar content.</p>
+    </aside>
+  {% endblock %}
 </div>
 
 <!-- Inline styles are for demo purposes only. Prefer external stylesheet in production. -->
 <style>
   .mzp-l-main {
-    background-color: papayawhip;
+    border: 3px solid papayawhip;
```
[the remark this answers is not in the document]
maureenlholland: Good catch! Yes, I switched from background colors because it looked too busy on the sidebar with menu example, but, while they were good as subtle background colors, they are definitely too light for borders. Will update.

components/main-with-sidebar/main-with-sidebar--sidebar-menu.html
maureenlholland: @craigcook let me know what you think of the general structure here. I wasn't able to pull the example sidebar menu out without grouping into nested folders. Is this an improvement? Or do we want to fallback to the pseudo-variant sidebar menu example for now?

menu-nesting
folder-nesting

craigcook: LGTM. Numbering filenames and folders is how Fractal explicitly tells you to order things (Drizzle would let us define the order in a config).

components/main-with-sidebar/example-sidebar-menu/example-sidebar-menu.html
```diff
@@ -0,0 +1,12 @@
+{% extends '@main-with-sidebar--default' %}
+
+{% block content %}
+  <aside class="mzp-l-sidebar">
+    {% render '@sidebar-menu' %}
+  </aside>
+  <main class="mzp-l-main">
+    <p>Main content.</p>
```
craigcook, line 8: In the interest of making this a more realistic example we should fill this with more dummy text. A few paragraphs of lorem ipsum, some headings, etc. And we can remove the colored borders so it's a more accurate presentation. Not sure how best to do that with `extends`, maybe just put the styles in a block in the parent template so this one can have an empty block.

components/main-with-sidebar/01-main-with-sidebar/readme.md
```diff
@@ -0,0 +1,17 @@
+A block with a main column and one sidebar on either the left or right (this example uses `mzp-l-sidebar-left`). A single page should only have one main section and one sidebar.
```
craigcook, line 1: We should mention in the notes that the outlines are just for demonstration, they're not part of the component. Lest someone assume it comes with colored borders.

craigcook (changes requested):

craigcook (approved):

PR #792, opened 2022-06-09
Make sure to build package using production mode

alexgibson:
## Description

I'm not sure this technically makes any difference to our generated code, since we're manually creating both minified and non-minified JS files, but we should probably still set`--mode production` explicitly whenever we build the NPM package.

### Issue

N/A

### Testing
- [ ] `npm test` should build the package and successfully run tests against the generated files.
- [ ] Generated files should still look the same.


craigcook (approved):
💻

PR #793, opened 2022-06-09
Migrate Cards and Feature Card

craigcook:
## Description
Migrates Card and Feature Card components.

### Testing
http://localhost:3000/components/detail/card
http://localhost:3000/components/detail/feature-card--default


components/card/card.html
```diff
@@ -0,0 +1,27 @@
...
+# meta          - string
+#}
+<section class="mzp-c-card {{ class }}">
+  <a class="mzp-c-card-block-link" href="{{ link }}">
+    <div class="mzp-c-card-media-wrapper">
+      <img class="mzp-c-card-image" src="{{ image | default('/img/image-3-2.jpg') | path }}" alt="">
+    </div>
+    <div class="mzp-c-card-content">
+      {% if tag %}<div class="mzp-c-card-tag">{{ tag }}</div>{% endif %}
+      {% if title %}<{{ title_el | default('h2') }} class="mzp-c-card-title">{{ title }}</{{ title_el | default('h2') }}>{% endif %}
```
reemhamz, line 21: **question:** What's `title_el`?

components/card/card.html
```diff
@@ -0,0 +1,27 @@
+{# @card
```
reemhamz, line 1: Few things about this component:
- The pattern of displaying the page seems a bit odd here. When I click `Card` from the sidemenu, it takes me to `http://localhost:3000/components/detail/card`, which I then can't access back if I click on one of the variants, and I'd have to click on `Card` again to see this page. Seems like an anti-pattern to what we've been structuring so far? I would have assumed this page, that showcases all the different cards, to probably be an `All cards example` page or something, similar to [what you did here](https://demo1--mozilla-protocol.netlify.app/components/detail/example-form.html). What do you think?
- The hierarchy of displaying the variants also could be in an ascending or descending order (like we have in https://protocol.mozilla.org/patterns/molecules/card.html that displays the Large card first, and descends to the Extra Small card), as it would be easier to navigate? I think you chose Small to be the default/first item in the list because it doesn't need an extra sizing class, but I think it's still better to have a clearer hierarchy on the list?
- I also see you added the `preview-dark.html` file in this PR to probably show a dark variant of the card. It would be nice to also include that if you could?

components/card/card--small.readme.md
```diff
@@ -0,0 +1,5 @@
+- Small sized cards can contain images with 16:9, 3:2 or 1:1 aspect ratios.
```
reemhamz, line 1: **Suggestion (non-blocking):** could we probably move these notes into the `card.config.yml` file? There is already a `readme.md` file for when you click on the Card item from the sidemenu, and it would probably be more neat to have these notes also within the YAML file?

reemhamz (changes requested):
Good work on this!
I left a few comments that we should discuss before merging this that are important to bring up imho. 🪂

components/card/card.html
craigcook: Yeah, this got a little weird... Small is the default, but I wanted some separate "overview" documentation not attached to a variant. So I used the collated: true setting which tells Fractal to put all the variants on one page... but then you can't easily get back to that overview. So I'm not sure I like collated components and maybe we should avoid that.

I've updated to add a "Default" and treat "Small" as a variant, even if they're more or less the same thing, but that it least still keeps things orderly.

I think we tend to list sizes from small to large so I was just trying to be consistent. If we prefer listing from large to small we should do that everywhere. I'm not sure I have a strong opinion on which, but we should pick one pattern and stick to it.

The dark preview already existed, I just made a minor edit to it in this PR. I did have a dark card at one point but decided to remove it.

reemhamz: Happy to stick to ascending order (small to large).
I do however think you should either add a variant to show the dark card, or to at least include in the notes section that you can add an `mzp-t-dark` class for cards

reemhamz (approved):
LGTM! r+wc
Just a quick note on either adding a dark variant to cards or to say that could add the `mzp-t-dark` class in the notes section for cards.

PR #802, opened 2022-06-16
Add direction toggle

craigcook:
## Description
Adds a LTR/RTL toggle to the theme to change the text direction of the component preview.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,9 +1,10 @@
 module.exports = {
     env: {
-        'node': true,
+        'browser': true,
         'commonjs': true,
+        'es2017': true,
         'jasmine': true,
-        'es2017': true
+        'node': true
     },
     extends: [
         'eslint:recommended'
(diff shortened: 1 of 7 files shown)


craigcook:
This is only semi-functional at the moment. It works when you do a page refresh but not when you're navigating around the set because clicking links does some ajaxy stuff to change the content rather than actually loading the page, which means the toggle doesn't init. 

I think we may be able to piggy-back on the event that changes content around to init the toggle but that's beyond my skills with JS.

alexgibson:
I opened a PR to your branch to fix the toggle

craigcook:
Note, this also removes the padding from the component preview. I don't like how it looks without that bit of spacing around smaller components like single buttons, but it gives an unrealistic presentation of full-width components like Split. I think I may add a separate padded preview to use when needed but for now let's lose the spacing.

alexgibson (approved):
r+

PR #804, opened 2022-06-17
Migrate newsletter form

craigcook:
## Description
Adds the newsletter form component. 

Honestly I put minimal effort into migrating this one, it's very much copy-paste. This component is long overdue for a redesign and/or just being deprecated. It was designed very specifically for mozilla.org and a specific purpose like four years ago and it just feels out of place in Protocol now. But it exists so it should be documented, we just don't need to overthink it.

### Testing
http://localhost:3000/components/detail/newsletter


reemhamz (approved):
r+ 🦍

components/newsletter/newsletter.html
```diff
@@ -0,0 +1,56 @@
+<aside class="mzp-c-newsletter">
+  <div class="mzp-c-newsletter-image">
+    <img src="{{ '/img/newsletter-image.png' | path }}" width="346" height="346" alt="">
+  </div>
+
+  <form id="newsletter-form" class="mzp-c-newsletter-form" name="newsletter-form" action="#" method="post">
```
reemhamz, line 6: Gg with the addition of the  variants, I think this is a nice way of showing things!

However, still getting something like this in the initial form submit.
Screen Shot 2022-06-22 at 10 15 05 AM
Nothing blocking, but we could remove `method="post"` and change it to `method=""` to prevent it from trying to submit and we could leave a comment in the notes saying there is no `method` attribute just for demonstration purposes. 

Again, nothing blocking, just less error-y would be cool.

craigcook: oh good idea

reemhamz (approved):
r+ thanks for the changes!

PR #807, opened 2022-06-28
Update contributing doc, add usage docs, general doc tidying

craigcook:
It's a fair bit of content but not much to actually test. It just needs a proofread for spelling/grammar and to make sure everything makes sense and is accurate.


reemhamz (approved):
A few suggestions but otherwise is great. Nice job! r+wc 🦞

Issue #800, opened 2022-06-16
Wrap pixel values in `code` syntax in Fractal migration


reemhamz:
## Description
Based on [this conversation](https://github.com/mozilla/protocol/pull/786#discussion_r894131785), we're going to wrap pixel values in the Fractal migration inside backticks for code-like syntax. This can help users quickly navigate to pixel values they'd need to use.

We can test this out for a bit and revert if we don't like it much, however I tend to see code when I read pixel values, and would be a nice and neat implementation.

craigcook:
I think once we have all the components added in I'm going to make a "consistency pass" through everything to look for tweaks to formatting, indentation, variable names, etc.

reemhamz:
Good idea. This can wait till migration is finished, then

PR #808, opened 2022-06-30
Tidy component code and docs, improve consistency

craigcook:
## Description
Taking a pass through all the component files to give them consistent formatting. 

* Wrapped lines at 80 columns where possible.
* Use curly quotes and apostrophes in documentation.
* Use consistent variable names (like `title` instead of `heading`). 
* Under_scores instead of camelCase.
* Added reference comments in a few components that were lacking them. 
* Update links to components.
* Wrap px units in code ticks (#800)

It looks like a lot because it touches so many files but the actual changes are pretty superficial.

### Testing
Just look everything over and make sure nothing is broken and that hopefully I didn't miss anything.


maureenlholland (changes requested):
Looks a lot cleaner, nice work 🧹 ✨ 
Minor changes

A couple of files that seem to be missing parameter comments:
- menu-list
- modal
- navigation
- menu-item
- sticky-promo
- split
- picto-card

A couple of pixel backticks missing config file notes.
- card (450px)
- hero (800px)
- picto (64px)
(I think most of these are already suggested changes)

Sidenote: we might want to consider a config to enforce max line length and single quotes (I think editorconfig can do a max line length, but might need prettier for quotes)

PR #816, opened 2022-07-14
Added accessible attributes to menu bar (fixes #815)

### LINKED ISSUE #815: Improve accessibility with the new Fractal migration
## Description
Accessibility with the Fractal site is at an 80% based on Lighthouse's score:
Screen Shot 2022-07-14 at 11 19 46 AM

Most of the accessibility issues are coming from the menu bar, which should be easy enough to fix.

## Steps to reproduce
- Open Chrome browser
- Open dev tools
- Select `Lighthouse`
- Click `analyze page load`

I'm leaving a screenshot to give a visual of things to fix:

Screen Shot 2022-07-14 at 11 22 13 AM
Screen Shot 2022-07-14 at 11 22 35 AM

reemhamz:
## Description
Some areas in the menu bar lacked accessibility, notably a lack of an `alt` for the Github icon and a lack of a `role` for the nav toggle. Just patched that up!
There are still some accessibility issues that should be patched up, but it's not easy for us to pull that since they're coded within Fractal's codebase and dependencies, notably:
- The nav headers are defaulted to be `h3` but since there is no `h2` in the DOM all the time (depending on which page you're viewing), the hierarchy isn't very optimal
- The `iframe` tag isn't supposed to have a `title` attribute, mainly seen in the preview pages

### Improvements:
Before this PR, accessibility was at 80% in Lighthouse, now it's at 98%:
Screen Shot 2022-07-14 at 5 12 17 PM

### Testing
I used MacOS' VoiceOver tool to see if it went through the menu's DOM reading correctly

### Issue
https://github.com/mozilla/protocol/issues/815

### DIFF
--- a/theme/views/partials/header.nunj
+++ b/theme/views/partials/header.nunj
@@ -1,5 +1,5 @@
 <div class="Header">
-  <button class="Header-button Header-navToggle" data-action="toggle-sidebar">
+  <button class="Header-button Header-navToggle" data-action="toggle-sidebar" aria-label="menu">
       <div class="Header-navToggleIcon Header-navToggleIcon--open">
           {% include 'icons/close.svg' %}
       </div>
@@ -20,7 +20,7 @@
   {% endif %}
     <li>
       <a href="https://github.com/mozilla/protocol" title="Protocol on Github">
-        <img src="{{ path('/theme/img/github.svg') }}" height="30">
+        <img src="{{ path('/theme/img/github.svg') }}" height="30" alt="">
       </a>
     </li>
   </ul>
--- a/theme/views/partials/header.nunj
+++ b/theme/views/partials/header.nunj
@@ -1,5 +1,5 @@
 <div class="Header">
-  <button class="Header-button Header-navToggle" data-action="toggle-sidebar" aria-label="Toggle menu">
+  <button class="Header-button Header-navToggle" data-action="toggle-sidebar" aria-label="Toggle menu" aria-controls="navigation">
       <div class="Header-navToggleIcon Header-navToggleIcon--open">
           {% include 'icons/close.svg'%}
       </div>
--- a/theme/views/partials/header.nunj
+++ b/theme/views/partials/header.nunj
@@ -1,7 +1,7 @@
 <div class="Header">
-  <button class="Header-button Header-navToggle" data-action="toggle-sidebar" aria-label="menu">
+  <button class="Header-button Header-navToggle" data-action="toggle-sidebar" aria-label="Toggle menu">
       <div class="Header-navToggleIcon Header-navToggleIcon--open">
-          {% include 'icons/close.svg' %}
+          {% include 'icons/close.svg'%}
       </div>
       <div class="Header-navToggleIcon Header-navToggleIcon--closed">
           {% include 'icons/burger.svg' %}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 ## Bug Fixes
 * **node:** Make sure to build NPM package using production mode.
-* **html:** Added accessible attributes to menu bar
+* **html:** Added accessible attributes to menu bar (#815).
+
 # 16.0.1
 
 ## Bug Fixes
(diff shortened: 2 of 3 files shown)


theme/views/partials/header.nunj
```diff
@@ -1,5 +1,5 @@
 <div class="Header">
-  <button class="Header-button Header-navToggle" data-action="toggle-sidebar">
+  <button class="Header-button Header-navToggle" data-action="toggle-sidebar" aria-label="menu">
```
nathan-barrett, line 2: **Question**: if this is a toggle for the menu - did you want to change this to a `aria-role`? or change what the label since this is the toggle and not the menu itself?

nathan-barrett (approved):
r+wc

theme/views/partials/header.nunj
reemhamz: Thanks for the comment, it actually got me thinking:
This button shouldn't have `aria-label`/`role` of `menu`, since it's not directly a part of the menu, it just has an action that toggles the menu, which is not part of the same exact tree branch, if you get what I mean.

Now I'm thinking of just giving it `aria-label="toggle menu"` because it's just a toggle to close/open the menu, and that should be clear for screen readers. I do want to also place `aria-label` on both the close toggle and open toggle, however they are `div`s (see here: https://github.com/mozilla/protocol/blob/main/theme/views/partials/header.nunj#L3), and according to this [accessibility site](https://www.davidmacd.com/blog/does-aria-label-override-static-text.html), adding `aria-label` on a `div` will **only** work if it also comes with a `role` attribute. However, I'm not sure which role I should give the `div` in this case if its just holding an icon? Maybe it should be `role="img"`, but I don't think that would be correct in this case, happy to hear what other devs have to say about this.

alexgibson: The most semantic solution here would just be `<button>Toggle menu</button>` instead of fiddling around with aria attributes, however that would require a larger refactor of the button, which may not be possible given ithe current markup is something Fractal could rely on.

So, instead I think using `aria-label="Toggle menu"` makes sense to me @reemhamz. However, I'm not sure you really need to go further with adding additional attributes to the `<div>`'s, as these are child elements of the `<button>` which has the required semantics?

Adding an `aria-controls` attribute to the menu icon could be an additional improvement to help screen readers understand what the button does when clicked.

reemhamz: Thanks for your insight Alex!
I wanted to add additional aria attributes to the `div`'s because one was to close the menu, and the other to open it, and each option had a respective icon to go with it. 

I don't think, depending on the hierarchy of how the button is placed several layers prior to the actual `nav` element, that it would make a difference here to screenreaders, especially considering that even if the sidemenu isn't showing, the screenreader is still going through the options. 
I suppose this is something from Fractal's end, and we'd have to do more refactoring if we want it to be optimal.
For now, I'll stick to adding the `aria-label="Toggle menu"` attribute on the button element.

alexgibson:
> Would you happen to know how we can add an id attribute to the nav here? Much of it seems hard-coded?

@reemhamz looking at the Fractal docs, it seems like we can override any nunjucks template to modify things we might need, such as adding an `id`.

It looks like this is the template source: https://github.com/frctl/fractal/blob/main/packages/mandelbrot/views/partials/navigation/navigation.nunj

If you make a copy of this and then add it to `./theme/views/partials/navigation/navigation.nunj` then you can add an `id` to it. Seems to work for me locally.

reemhamz:
Thanks @alexgibson, that seemed to work 🥳.

I did more research on `aria-controls`, and though it definitely is beneficial, [not many screen-readers support it](https://a11ysupport.io/tech/aria/aria-controls_attribute). I'm going to add it in for now because it can at least help a user using something like JAWS that has support for it, but this is definitely something we should be wary of regarding screen-readers (for front-end education sake).

alexgibson:
JAWS is widely used on Windows, so I think it’s a worthwhile thing to add 👍

reemhamz:
Let me know if it's good to merge with the changes 💯

Issue #814, opened 2022-07-12
Update .stylelintrc to match config from 'mozilla/bedrock'


nathan-barrett:
## Description

Update the existing `.stylelintrc` file to extend the `stylelint-config-standard-scss` rules and match the config file in `mozilla/bedrock` [here](https://github.com/mozilla/bedrock/blob/main/.stylelintrc)

### Success Criteria

- [ ] `.stylelintrc` has been updated 
- [ ]  affected files have been reformatted

nathan-barrett:
@katiechurchwell assigned! A couple notes I forgot to mention:
- [ ] the library `stylelint-config-standard-scss` will need to be added via npm
- [ ] run `npm run lint-css` to lint all `.scss` files before your PR so that the files have been reformatted correctly

katiechurchwell:
@nathan-barrett Ok, I have a quick update and a question for you!

 - :heavy_check_mark: Extended and added the dependency for `stylelint-config-standard-scss`
- Partially reformatted when running `npx stylelint "**/*.scss" --fix`
  - `npm run lint-css` doesn't work for me which makes me feel like the script should be updated?
  - getting a lot of `no-descending-specificity` messages so figuring out what that means. I think it has to do with how the custom rules are written?
- When you say "match the config file in `mozilla/bedrock`", do you want the file to look exactly like that or just the rules?

Thanks! :cowboy_hat_face:

nathan-barrett:
Cool! Nice work @katiechurchwell - I think I miscommunicated what the best way to do this - I think that the `.stylelintrc` file in protocol should contain the same rules that are found inside the bedrock repo - but also keep the existing rules that are in place in protocol.  If there are two matching and/or conflicting rules - we should use the preference set in bedrock.  

Let me know if any of that doesn't make sense.

As for the `no-descending-specificity` messages - its coming from this rule [here](https://stylelint.io/user-guide/rules/list/no-descending-specificity/).

katiechurchwell:
@nathan-barrett Hi Nathan, I wanted to provide an update on this issue. It's coming along but I'm concerned about some of the linting rules causing some breaking changes; kebab case and no global function names in particular. I could go ahead and format everything and submit a PR to have you check it out but I wanted to doublecheck these are the intended changes.

nathan-barrett:
@katiechurchwell good call out! Lets keep the duplicated variable names. 

[...]

PR #820, opened 2022-07-27
Style lint

katiechurchwell:
## Description

Brief summary:
- SCSS linting update to match `mozilla/bedrock`.

### Issue

[Update .stylelintrc to match config from 'mozilla/bedrock'.](https://github.com/mozilla/protocol/issues/814)


nathan-barrett:
hiya @katiechurchwell!

Found the following rules that can be removed from the .stylelintrc.json since they are already the same when using `stylelint-config-standard-scss` (looked [here](https://github.com/stylelint/stylelint-config-recommended/blob/main/index.js) for reference):
- `color-no-invalid-hex`
- `font-family-no-duplicate-names`
- `unit-no-unknown`
- `property-no-unknown`
- `keyframe-declaration-no-important`
- `declaration-block-no-shorthand-property-overrides`
- `block-no-empty`
- `selector-pseudo-class-no-unknown`
- `selector-pseudo-element-no-unknown`
- `selector-type-no-unknown`
- `comment-no-empty`
- `no-invalid-double-slash-comments`

assets/sass/protocol/base/elements/_forms.scss
```diff
@@ -1,12 +1,12 @@
+/* stylelint-disable declaration-colon-newline-after */
```
craigcook, line 1: Does this need to be at the top of the file? It looks like the same feature is disabled for a line below so this might be redundant. If we do need a disable comment at the top it should probably go after the license block.

assets/sass/protocol/base/elements/_forms.scss
```diff
@@ -42,79 +49,73 @@ textarea {
...
+[type='search'] {
+    appearance: textfield; // 1
     outline-offset: -2px; // 2
 }
 
 // Remove the inner padding in Chrome and Safari on macOS.
-[type="search"]::-webkit-search-decoration {
-    -webkit-appearance: none;
+[type='search']::-webkit-search-decoration {
+    appearance: none;
```
craigcook, line 73: Some of these rules are specifically for older versions of Safari, so we might still need the prefixed property when it's also using a webkit-only selector. I could be wrong, we should probably research these a bit.

.stylelintrc.json
```diff
@@ -1,6 +1,44 @@
...
+            {
+                "camelCaseSvgKeywords": true,
+                "ignoreFunctions": ["local"]
+            }
+        ],
+        "max-line-length": null,
+        "font-family-no-missing-generic-family-keyword": [
+            true,
+            {
+                "ignoreFontFamilies": ["FA-Icons-Contact"]
```
craigcook, line 27: We don't have this font in Protocol (I'm not even sure we still have it in bedrock) so we don't need this exception.

craigcook (changes requested):
Just a few drive-by comments. I'm not sure about some of the webkit-specific form stuff, we might still need those prefixes. If so, we can add disabling comments to keep stylelint happy.

assets/sass/protocol/base/elements/_forms.scss
nathan-barrett: for this one lets add a comment on the line above to disable the particular rule since its a bit of a one-off:

```css

 /* stylelint-disable-next-line value-no-vendor-prefix */
 -webkit-appearance: none;

```

assets/sass/protocol/base/elements/_forms.scss
```diff
@@ -42,79 +49,73 @@ textarea {
...
 }
 
 // 1. Correct the inability to style clickable types in iOS and Safari.
 // 2. Change font properties to `inherit` in Safari.
 ::-webkit-file-upload-button {
-    -webkit-appearance: button; // 1
+    appearance: button; // 1
```
nathan-barrett, line 79: @katiechurchwell basically anwhere we see `-webkit-appearance` we should update with a ignore comment on the line previous

nathan-barrett:
@katiechurchwell sorry for making this super, extra complicated, but lets update the stylelint to this: (it removes duplicates which are extended from `stylelint-config-standard-scss`)

```json
{
    "extends": "stylelint-config-standard-scss",
    "rules": {
        "indentation": 4,
        "string-quotes": "single",
        "declaration-empty-line-before": "never",
        "declaration-no-important": true,
        "color-function-notation": "legacy",
        "alpha-value-notation": "number",
        "property-no-vendor-prefix": null,
        "value-no-vendor-prefix": null,
        "at-rule-empty-line-before": null,
        "no-descending-specificity": null,
        "value-keyword-case": [
            "lower",
            {
                "camelCaseSvgKeywords": true,
                "ignoreFunctions": ["local"]
            }
        ],
        "max-line-length": null,
        "declaration-block-no-redundant-longhand-properties": [
            true,
            {
                "ignoreShorthands": [
                    "grid-area",
                    "grid-column",
                    "grid-gap",
                    "grid-row",
                    "grid-template"
                ]
            }
        ],
        "function-url-no-scheme-relative": true,
        "max-nesting-depth": 5,
        "no-unknown-animations": true
    }
}
```

craigcook (approved):

assets/sass/protocol/includes/_mixins.scss
```diff
@@ -168,43 +166,41 @@
...
-              column-gap: $gap;
+    column-width: $width;
+    column-width: $width;
+    column-width: $width;
+    column-count: $count;
+    column-count: $count;
+    column-count: $count;
+    column-gap: $gap;
+    column-gap: $gap;
+    column-gap: $gap;
```
alexgibson, line 203: Nit: looks like the linter created some duplicate properties here, can these now be trimmed down?

assets/sass/protocol/includes/_mixins.scss
```diff
@@ -27,112 +27,110 @@
...
+    box-decoration-break: $value;
 }
 
 @mixin appearance($value) {
-    -webkit-appearance: $value;
-       -moz-appearance: $value;
-	        appearance: $value;
+    appearance: $value;
+    appearance: $value;
+    appearance: $value;
```
alexgibson, line 131: duplicate properties

assets/sass/protocol/includes/_mixins.scss
```diff
@@ -27,112 +27,110 @@
...
-            background-size: $sizes;
+    background-size: $sizes;
+    background-size: $sizes;
 }
 
 @mixin box-decoration-break($value: slice) {
-    -webkit-box-decoration-break: $value;
-            box-decoration-break: $value;
+    box-decoration-break: $value;
+    box-decoration-break: $value;
```
alexgibson, line 125: duplicate properties

assets/sass/protocol/includes/_mixins.scss
```diff
@@ -27,112 +27,110 @@
...
     -webkit-box-pack: $justify;
-     justify-content: $justify;
+    justify-content: $justify;
 }
 
 @mixin background-size($sizes...) {
-    -webkit-background-size: $sizes;
-            background-size: $sizes;
+    background-size: $sizes;
+    background-size: $sizes;
```
alexgibson, line 120: duplicate properties

assets/sass/protocol/includes/_mixins.scss
```diff
@@ -27,112 +27,110 @@
...
     -webkit-box-direction: $value;
-           flex-direction: $value;
+    flex-direction: $value;
 }
 
 @mixin flex-wrap($value: wrap) {
-    -ms-flex-wrap: $value;
-        flex-wrap: $value;
+    flex-wrap: $value;
+    flex-wrap: $value;
```
alexgibson, line 105: duplicate properties

assets/sass/protocol/includes/_mixins.scss
```diff
@@ -27,112 +27,110 @@
...
 }
 
 @mixin flex($values: auto) {
     -webkit-box-flex: $values;
-       -moz-box-flex: $values;
-        -webkit-flex: $values;
-                flex: $values;
+    -moz-box-flex: $values;
+    flex: $values;
+    flex: $values;
```
alexgibson, line 95: duplicate properties

assets/sass/protocol/includes/_mixins.scss
```diff
@@ -27,112 +27,110 @@
...
 @mixin flexbox {
-    display: -webkit-box;
-    display: -moz-box;
-    display: -ms-flexbox;
-    display: -webkit-flex;
+    display: box;
+    display: box;
+    display: flexbox;
+    display: flex;
     display: flex;
```
alexgibson, line 88: duplicate properties

assets/sass/protocol/includes/_mixins.scss
```diff
@@ -2,8 +2,9 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-@use "sass:math";
+/* stylelint-disable declaration-block-no-duplicate-properties */
```
alexgibson, line 5: Ok, there are a lot of duplicate properties in this file. If we fix those, can this comment then be removed? 😅

assets/sass/protocol/includes/mixins/_gradients.scss
```diff
@@ -1,54 +1,54 @@
+/* stylelint-disable declaration-block-no-duplicate-properties */
```
alexgibson, line 1: Removing the `-webkit` vendor prefixes for `background-image` in this file has created duplicate properties, but it looks like only the second occurrence (the standard-based rule) is now valid.

If we fix the duplicates here, can we remove this comment?

assets/sass/protocol/components/logos/_wordmark-product-vpn.scss
```diff
@@ -1,14 +1,14 @@
+/* stylelint-disable scss/dollar-variable-pattern */
```
alexgibson, line 1: Nit: some of these comments are hard to spot when they're bunched right up to the licence block below. Suggest placing these on a separate line underneath the licence block for better readability?

(not a blocker)

alexgibson (changes requested):
I spotted a (possible) issue caused by duplicate properties triggered by vendor prefix removal in the two files below:

- `/protocol/assets/sass/protocol/includes/_mixins.scss`
- `/protocol/assets/sass/protocol/includes/mixins/_gradients.scss`

It looks like we (might?) be able to remove `/* stylelint-disable declaration-block-no-duplicate-properties */` from those files if we fix the issues.

Other than that, I didn't spot any immediate issues or errors.

Nice work and thanks for the PR!

assets/sass/protocol/includes/mixins/_gradients.scss
```diff
@@ -1,54 +1,54 @@
...
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-
 // Horizontal gradient, from left to right.
 // Creates two color stops, start and end, by specifying a color and position for each color stop.
 @mixin gradient-horizontal($start-color: #555, $start-percent: 0%, $end-color: #333, $end-percent: 100%) {
-  background-image: -webkit-linear-gradient(left, $start-color $start-percent, $end-color $end-percent); // Safari 5.1-6, Chrome 10+
-  background-image:         linear-gradient(to right, $start-color $start-percent, $end-color $end-percent); // Standard, IE10, Firefox 16+, Opera 12.10+, Safari 7+, Chrome 26+
+    background-image: linear-gradient(to left, $start-color $start-percent, $end-color $end-percent); // Safari 5.1-6, Chrome 10+
+    background-image: linear-gradient(to right, $start-color $start-percent, $end-color $end-percent); // Standard, IE10, Firefox 16+, Opera 12.10+, Safari 7+, Chrome 26+
```
alexgibson, line 10: Here's an example of where we have duplicate `background-image` properties, but the first property is using an invalid (non-standards based) syntax, because it used to be prefixed with `-webkit`. I think we can remove the first property and just keep the second.

There are a few more occurrences below

katiechurchwell:
> I spotted a (possible) issue with duplicate properties triggered by vendor prefix removal in the two files below:
> 
> * `/protocol/assets/sass/protocol/includes/_mixins.scss`
> * `/protocol/assets/sass/protocol/includes/mixins/_gradients.scss`
> 
> It looks like we (might?) be able to remove `/* stylelint-disable declaration-block-no-duplicate-properties */` from those files if we fix the issues.
> 
> Other than that, I didn't spot any immediate errors.
> 
> Nice work and thanks for the PR!

Thank you for the review and thanks to @nathan-barrett for resolving those issues! :)

assets/sass/protocol/components/_button.scss
```diff
@@ -2,8 +2,6 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-/* stylelint-disable no-duplicate-selectors */
```
alexgibson: Comment (non-blocking): FWIW, I think it's probably OK to have rules like this at the top of a file, if the whole file relies upon it. My nit in the prior review was that it shouldn't come on line 1, above the license block, as that makes it hard to spot since eyes can misinterpret it as a singular comment.

But no need to change it again as what you have here still works :)

assets/sass/protocol/includes/_mixins.scss
```diff
@@ -30,69 +28,53 @@
 // Mixins for CSS3 properties that still need prefixes in some browsers.
 @mixin transition($transition...) {
     transition: $transition;
-    transition: $transition;
 }
 
 @mixin transform($transform...) {
     transform: $transform;
-    transform: $transform;
 }
```
alexgibson, line 35: Comment (non-blocking): it occurs to me that if we're removing vendor prefixes, then these mixins are all essentially now redundant? Not something we need to remove here and now, but we should probably file a follow up issue to remove them (which would also be a breaking change for Protocol).

assets/sass/protocol/includes/_mixins.scss
```diff
@@ -30,69 +28,53 @@
...
-    display: flex;
     display: flex;
 }
 
 @mixin flex($values: auto) {
     -webkit-box-flex: $values;
     -moz-box-flex: $values;
     flex: $values;
-    flex: $values;
 }
```
alexgibson, line 78: Issue (possibly blocking?) should these vendor prefixes also be removed? It's not quite clear to me why these are still here if they have been removed elsewhere.

alexgibson (approved):
A couple of open questions and one minor issue, but otherwise this looks good to me.

r+wc

assets/sass/protocol/includes/_mixins.scss
nathan-barrett: Yeah, I'll file an issue and we can talk over what this looks like a little more.

PR #827, opened 2022-10-24
Switch from US-only MM/DD/YYYY date format to the unambiguous 'MMM DD, YYYY'

globau:
## Description

The 'Built on' date at the bottom of the sidebar uses the US-only date
format of MM/DD/YYYY.  This date format is ambiguous and is incorrect
for most of the world.

Fix this by replacing the month with its short name.
eg. `10/25/2022` becomes `Oct 25, 2022`.


craigcook (approved):
📅

PR #831, opened 2022-11-08
Remove aria-disabled from disabled input docs and example (fixes #823)

### LINKED ISSUE #823: Input not needing aria-disabled?
From multiple sources I read, `input disabled` seems to already covered the function of `aria-disabled`. But the current doc suggests to add both:

https://protocol.mozilla.org/components/detail/input--disabled.html

Is there a reason behind it? Or can we remove that

maureenlholland:
## Description

Removes `aria-disabled` from disabled input docs notes and HTML. It is not applicable for the use case we are demonstrating. `disabled` attribute only is what we want.

### Issue

https://github.com/mozilla/protocol/issues/823

### Testing

`npm run webpack-docs`
(in another terminal tab, same path) `npm start`

http://localhost:3000/components/detail/input--disabled

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * **html:** Added accessible attributes to menu bar (#815).
 * **css:** Add style rule for the hidden attribute in global reset (#783).
 * **html:** Use unambiguous date format in sidebar.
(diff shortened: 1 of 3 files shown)


nathan-barrett (approved):
r+

Issue #812, opened 2022-07-10
Previous URLs are broken with new docs


maxxcrawford:
## Description

I have a few specific pages bookmarked from the Protocol docs site. (https://protocol.mozilla.org/fundamentals/tokens.html) 

### Steps to reproduce

- Go to https://protocol.mozilla.org/fundamentals/tokens.html
- Expected: Page loads correct "tokens" page
- Actual: 404 Error. 

### Expected result

- The old URLs redirect to the new pages. (https://protocol.mozilla.org/docs/fundamentals/design-tokens.html) 

### Actual result

404 Page Not Found

### Environment

N/A

craigcook:
Fixed with https://github.com/mozilla/protocol/pull/846

PR #846, opened 2023-01-31
Update netlify.toml with redirects from old drizzle urls

nathan-barrett:
## Description

Updated netlfiy redirects from older drizzle URLs to the updated URLs for the fractal build

### Issue

 #812

### Testing
You can look at [demo2](https://demo2--mozilla-protocol.netlify.app/) to test this change:
**example**: https://demo2--mozilla-protocol.netlify.app/patterns/organisms/billboard -> https://demo2--mozilla-protocol.netlify.app/components/detail/billboard

This will also work with or without a trailing `.html` on the url

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -3,25 +3,25 @@
   command = "npm run build-docs"
 
 # redirects for older pages:
-# Fundementals
+# Fundamentals
 [[redirects]]
-  from = "/fundementals/principles*"
+  from = "/fundamentals/principles*"
   to = "/docs/fundamentals/principles"
 
 [[redirects]]
-  from = "/fundementals/themes*"
+  from = "/fundamentals/themes*"
   to = "/docs/fundamentals/brand-themes"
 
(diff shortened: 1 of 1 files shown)


netlify.toml
```diff
@@ -1,3 +1,201 @@
 [build]
   publish = "dist"
   command = "npm run build-docs"
+
+# redirects for older pages:
+# Fundementals
+[[redirects]]
+  from = "/fundementals/principles*"
```
craigcook, line 8: Typo: should be fund**a**mentals, here and below.

```suggestion
  from = "/fundamentals/principles*"
```

craigcook (changes requested):
Looks good! just the repeated typo to fix (hazards of copy/paste). I also wonder if it's worth redirecting each of the individual demo pages to home, but I'm happy to just let them 404 as well.

craigcook (approved):

Issue #848, opened 2023-02-03
Component Change Request: use AAA contrast standards on Error notification bar


maureenlholland:
### Please describe the change you wish to see

Aim for AAA WCAG color contrast standards on [Errors notification bars](https://protocol.mozilla.org/components/detail/notification-bar--error.html)

### Please describe why these changes need to be made

The current Error notification styles pass normal AA contrast checks but fail AAA normal text check
<img width="581" alt="current-contrast" src="https://user-images.githubusercontent.com/19650432/216604919-5fc97f36-1202-43b9-97cf-94fa51f63856.png">

In the case of Error notifications, it feels like we should go beyond normal requirements and make the text as easy to read as possible. Errors are already frustrating and it would compound user frustration to struggle to read the error description.
<img width="512" alt="error" src="https://user-images.githubusercontent.com/19650432/216605048-9cd4f771-763c-42a0-8e70-e5f225baf5f9.png">

### Please provide any designs or prototypes of the proposed changes

Enforcing large/bolded text might be the most effective solution. 

### Please describe where this component is currently in use

Screenshot above is from Newsletter error (locally forcing errors from the views.py file: http://localhost:8000/en-US/newsletter/confirm/thanks/?error=2)

### Please describe what this component should do in the following conditions

* When set in front of a dark background
* When set in front of a light background
* When viewed in a mobile/responsive viewport

I don't think it needs any changes to responsive viewport behaviour or dark/light theme

### Is the development of this component a blocking dependency for other work? Please explain if so

Not blocking

reemhamz:
Thanks for creating this issue! It's easy to realize that even though color contrast passes, our eyes may just generally need something a little better than what the green `PASS` stamp tells us.

I fiddled around with the error notification bar using dev tools. Two things I found most helpful (conducted separately):
- Changing font color to plain old fashion black (#000000)
![image](https://user-images.githubusercontent.com/42309026/217147215-618fdf98-5395-4e6d-ac73-7ed1aa977ce7.png)

- Increasing the font size to `16px`

![image](https://user-images.githubusercontent.com/42309026/217147506-af841017-8a82-4b8a-8118-093bd0316881.png)

Both seem to improve the notification bar (just increasing size by `2px` made such a difference for me), and I would prefer to add both of the changes. However, if we were only to choose one method, I would lean towards increasing the font-size rather than the color. 

[...]

PR #850, opened 2023-02-14
Darkened notification bar font color

reemhamz:
## Description

Related to issue #848, @maureenlholland and I spoke about how either increasing the `font-size` or  changing the `color` of the text in the notification bar would be slightly better for visual accessibility.
In the issue, I tested changes by:
- Increasing `font-size` to `16px`, and
- Changing `color` to `#000000`

Both seemed to be viable options for visual clarify on the component. However, a change in one of the notification bars would usually mean a change to the rest for sake of consistency. The increase in `font-size`, most notably on the success notification bar in the WNPs, seemed to jumble the order of the page's content hierarchy by just a bit. 

So instead of going with increasing `font-size` we opted to select to just change `color` to `#000000` for all notification bars. Not a significant change in itself, but still positively noticeable.

### Issue
#848 

### Testing
http://localhost:3000/components/detail/notification-bar--error (and the rest of the variants)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * **css:** Add style rule for the hidden attribute in global reset (#783).
 * **html:** Use unambiguous date format in sidebar.
 * **docs:** Remove aria-disabled from disabled form inputs.
+* **css:** Update font color for Notification Bar variants to black for better visual accessibility (#848).
 
 # 16.0.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 * **css:** Add style rule for the hidden attribute in global reset (#783).
 * **html:** Use unambiguous date format in sidebar.
 * **docs:** Remove aria-disabled from disabled form inputs.
-* **css:** Update font color for Notification Bar variants to black for better visual accessibility (#848).
+* **css:** Update font color for Notification Bar variants to black and error-variant background color to a lighter red for better visual accessibility (#848).
 
 # 16.0.1
 
--- a/assets/sass/protocol/components/_notification-bar.scss
+++ b/assets/sass/protocol/components/_notification-bar.scss
@@ -12,7 +12,7 @@
     border-radius: $border-radius-sm;
     border: 1px;
     box-shadow: $box-shadow-md;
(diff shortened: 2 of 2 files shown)


reemhamz:
@maureenlholland Carved this PR out for the sake of testing out the color changes, but it doesn't need to be final in case we still want to discuss this further. Let me know what you think!

ehot:
Screen Shot 2023-02-14 at 11 13 11 AM

I checked the colors that were in protocol and these were the results. They all passed; perhaps the "success button hover/active" state could be made lighter to try to get it to AAA, but other than that, I'm fine with these colors.

maureenlholland (approved):
r+ it's an improvement and passes AAA contrast check

personal preference from myself and @stevejalim is the lighter red background too, but that's our old eyes talking :)

reemhamz:
I'm comparing `$color-red-40` with `$color-red-30` and I prefer the lighter red, my eyes aren't as strained focusing on the font much (maybe it's the moving fatigue), so I'll change it to that and see how that goes :)

maureenlholland (approved):
current:
current
updated:
update

r++ I find the lighter red background and black font a significant improvement for readability

I don't think these notification bars are in (Mozilla only) [Figma Protocol components](https://www.figma.com/file/JKjmrE0EjVqMmKHcyCaTYB/Protocol-%22Project-Poutine%22---Style-Guide?node-id=1971%3A6310&t=TXyHdUidP5gYQBuM-0) yet, so we won't need an update there.

craigcook:
The original colors came directly from the brand guidelines, and even there it's noted that the error message only passes AA. 
image

I agree, lightening the background seems like the right choice. 

Though another option would be to darken the background and invert the text color. We use white on a dark red for inline error messages in forms and white on a not-as-dark red for error lists. (https://protocol.mozilla.org/components/detail/error-messages.html and https://protocol.mozilla.org/components/detail/error-list.html). I think we could revisit both of those and perhaps land on one standard for error states across the board.

reemhamz:
Thanks for all your feedback team!
I'm going to go ahead and merge this PR. We can revisit using a darkened red with a white font down the line and help to create a standard for error messages throughout Protocol.

Issue #545, opened 2020-02-26
Create a 404 page 


Errolyn:
We do not have a designed 404 page for the site. Lets make one :D

tikhsuP:
@Errolyn 
I would like to design it.
How is the 404 Page expected look?

### This is how the ERROR 404 Page at [https://protocol.mozilla.org/](https://protocol.mozilla.org/) looks like: 

![Screenshot (21)](https://user-images.githubusercontent.com/42881874/77258305-d78a8200-6c9f-11ea-8f8b-179d33658f64.png)

### This is how it looks like at [https://www.mozilla.org/en-US/](https://www.mozilla.org/en-US/):

![Screenshot (22)](https://user-images.githubusercontent.com/42881874/77258370-2b956680-6ca0-11ea-8be1-51bfe8d6f778.png)

Is the page expected to look similar to the latter for uniformity?

stephaniehobson:
Need a URL to point Netlify to.

Suggested content:
- include a link to file an issue for the page that 404ed.
- link to homepage.

reemhamz:
Closed w/ #858

PR #858, opened 2023-03-06
Add 404 page

reemhamz:
## Description

Added a `404` page to display when the URL doesn't match what we have set up.

### Issue
https://github.com/mozilla/protocol/issues/545

### Testing

- https://localhost:3000/docs/404

  - The redirect from `netlify.toml` can't be displayed on dev. I'm going to push this up to a demo server to mess around with broken links. 👇🏽
- https://demo2--mozilla-protocol.netlify.app/random-broken-link

### DIFF
--- a/docs/404.md
+++ b/docs/404.md
@@ -1,4 +1,10 @@
+---
+title: 404 error
+hidden: false
+---
+
 <div markdown="1" class="404-page">
+<div>
 <h1>Oops, we can't find this page</h1>
 <p>Sorry, sometimes URLs are broken!</p>
 </div>
@@ -7,4 +13,5 @@
   <li><p>Can't seem to find the page you're looking for? <a href="https://github.com/mozilla/protocol/issues">File an issue on Github</a> and we'll see if we can help you find it</p></li>
   <li><p>Curious about Mozilla? Visit our <a href="https://www.mozilla.org">website</a></p></li>
 </ul>
+</div>
 
--- a/docs/404.md
+++ b/docs/404.md
@@ -1,11 +1,11 @@
 ---
-title: 404 error
+title: Error 404
 hidden: false
 ---
 
 <div markdown="1" class="404-page">
 <div>
-<h1>Oops, we can't find this page</h1>
+<h2>Oops, we can't find this page</h2>
 <p>Sorry, sometimes URLs are broken!</p>
 </div>
 <ul>
@@ -14,4 +14,5 @@ hidden: false
   <li><p>Curious about Mozilla? Visit our <a href="https://www.mozilla.org">website</a></p></li>
 </ul>
 </div>
+
 
--- a/docs/404.md
+++ b/docs/404.md
@@ -4,15 +4,19 @@ hidden: false
 ---
 
 <div markdown="1" class="404-page">
-<div>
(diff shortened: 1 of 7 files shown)


netlify.toml
```diff
@@ -199,3 +199,13 @@
 [[redirects]]
   from = "/demos/*"
   to="/"
+
+[[redirects]]
+  from = "/broken-link*"
```
reemhamz, line 204: Just placed this URL here to test out the redirect and to see if it worked

theme/assets/sass/components/_404.scss
```diff
@@ -0,0 +1,11 @@
+/* stylelint-disable selector-class-pattern */
+.Tree-items {
+    .Tree-item {
+        // Hide 404 page from sidebar in production, since it has a different URL after Webpack build
+        a[href='/docs/404'],
```
maureenlholland, line 5: untested, but the [*=](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors) should select any `a` elements with `href` attribute containing `404` so it could simplify the selector list here
```suggestion
        a[href*='404'],
```

netlify.toml
```diff
@@ -2,6 +2,12 @@
   publish = "dist"
   command = "npm run build-docs"
 
+# 404 status
```
maureenlholland, line 5: **issue (blocking)**: this rule is catching everything before it has a chance to be redirected by one of the more specific rules below. Visiting https://protocol.mozilla.org/fundamentals/principles.html redirects to https://protocol.mozilla.org/docs/fundamentals/principles but https://demo2--mozilla-protocol.netlify.app/fundamentals/principles.html shows 404 instead of redirect

missing-redirect

**suggestion**: move wildcard 404 rule to the bottom so it only catches urls that don't match anything more specific

docs/404.md
```diff
@@ -0,0 +1,22 @@
+---
+title: Error 404
+hidden: false
+---
+
+<div markdown="1" class="404-page">
```
maureenlholland, line 6: **note (not blocking)**: if I add a third path segment, the styles won't load
test: https://demo2--mozilla-protocol.netlify.app/random/broken/link
missing-styles

I think this is an internal Fractal issue, given we pass the theme css to their config: https://github.com/mozilla/protocol/blob/main/theme/index.js#L14
but it's relative, so too many path segments breaks it

2 path segments, css uses `themes` folder
under-themes

3 path segments, css tries to find non-existent `random/themes` folder
under-random-themes

I'd say this particular bug is a `wontfix`, just noting it for future reference

maureenlholland (changes requested):
Nice work 🎉  Looks great on demo!

One blocking issue with the redirects rule order and one suggestion on CSS selectors, then this is good to merge r+wc

theme/assets/sass/components/_404.scss
reemhamz: TIL :)

netlify.toml
reemhamz: Thanks, eagle eye! 🦅 👁️

maureenlholland (approved):
r+ 💯

maureenlholland:
@reemhamz this is good to merge once changelog conflict is fixed

Issue #854, opened 2023-02-21
matchMedia.addListener is deprecated


alexgibson:
## Description

As per https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener, `addListener` is deprecated. Whilst most browsers alias it to the newer [`addEventListener`](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryListEvent), there's no guarantee that this will be true forever.

We should update where `addListener` is used in Protocol to use `addEventListener`, and then fallback to `addListener` for older browsers such as IE.

Rough feature detect (needs testing)

```javascript
if (window.matchMedia('all').addEventListener) {
    // use addEventListener('change')
else if (window.matchMedia('all').addListener) {
    // use addListener
}
```

reemhamz:
Been looking to get my hands dirty with JavaScript. Picking this 🎟 up

reemhamz:
Taking a look at caniuse.com, here are the comparisons between:
- [`addListener`](https://caniuse.com/?search=addlistener) vs 
- [`addEventListener`](https://caniuse.com/?search=addeventlistener)

Seems like `addEventListener` has even more support than `addListener`, and has been supporting IE for longer.
Based on that evaluation, I'm not sure we'd need the `addListener` fallback in our code?

alexgibson:
addEventListener is not supported on matchMedia in IE. If you try it, you should see an error

reemhamz:
Ah -- I missed that when I was reading up on the changes. Thanks Alex

reemhamz:
Closed with #859

PR #859, opened 2023-03-09
Move from depracated `addListener` to `addEventListener` function

reemhamz:
## Description

`MediaQueryList.addListener()` is deprecated and we should be replacing it with a more supported alternative, which is `EventTarget.addEventListener()`. 

Check this [compatibility list](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList#browser_compatibility) (`MediaQueryList` inherits `EventTarget`) for the browser versions were keeping the fall back for (i.e. < Safari 14)

**Things to note:**
I'm using a `change` event to detect there has been a resize, and I got that from reading this in MDN: https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/change_event

However, it should be noted that this event isn't listed in MDN's event references page: https://developer.mozilla.org/en-US/docs/Web/Events . I'll be filing an issue for this.

### Issue
https://github.com/mozilla/protocol/issues/854

### Testing
 I tested this on old browser versions as listed in the compatibility table I shared above:
- [x] Firefox version 54
- [x] Safari version 13
- [x] IE 10/11

### Testing URLs
- http://localhost:3000/components/detail/footer
- http://localhost:3000/components/detail/navigation--default
- http://localhost:3000/components/detail/menu

### DIFF
--- a/assets/js/protocol/protocol-footer.js
+++ b/assets/js/protocol/protocol-footer.js
@@ -32,7 +32,7 @@
                 }
 
                 if (window.matchMedia('all').addEventListener) {
-                    _mqWide.addEventListener('change', screenChange);
+                    _mqWide.addEventListener('change', screenChange, false);
                 } else if (window.matchMedia('all').addListener) {
                     _mqWide.addListener(screenChange);
                 }
--- a/assets/js/protocol/protocol-footer.js
+++ b/assets/js/protocol/protocol-footer.js
@@ -5,6 +5,15 @@
 (function() {
     'use strict';
 
+    // removes Details component if screen size is big
+    function screenChange(mq) {
+        if (mq.matches) {
+            Mzp.Details.init(footerHeadings);
+        } else {
+            Mzp.Details.destroy(footerHeadings);
+        }
+    }
+
     // check we have global variable
     if (typeof window.Mzp !== 'undefined') {
         var Mzp = window.Mzp;
@@ -14,7 +23,7 @@
         if (typeof Mzp.Supports !== 'undefined' && typeof Mzp.Details !== 'undefined') {
 
             // check browser supports matchMedia
-            if(Mzp.Supports.matchMedia) {
+            if (Mzp.Supports.matchMedia) {
                 var _mqWide = matchMedia('(max-width: 479px)');
 
                 // initialize details if screen is small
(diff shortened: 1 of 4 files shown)


assets/js/protocol/protocol-footer.js
```diff
@@ -14,22 +23,19 @@
...
-                _mqWide.addListener(function(mq) {
-                    if (mq.matches) {
-                        Mzp.Details.init(footerHeadings);
-                    } else {
-                        Mzp.Details.destroy(footerHeadings);
-                    }
-                });
+                if (window.matchMedia('all').addEventListener) {
+                    _mqWide.addEventListener('change', screenChange);
+                } else if (window.matchMedia('all').addListener) {
```
reemhamz, line 36: Doesn't look like we may need this fallback, since `addEventListener` has more support than `addListener`. Going to discuss with team

alexgibson: The fallback is needed for IE I believe

reemhamz: **Note**: Turns out I missed looking up the `matchMedia` method during research. Since that method returns a `MediaQueryList` object, it should be noted that `MediaQueryList.addEventListener` doesn't work in Safari and IE, as mentioned here under Known Issues: https://caniuse.com/?search=matchMedia, so a fallback will be necessary as Alex pointed out above.

reemhamz:
Looks like our `addListener` in prod event doesn't work on older Chrome browsers (< version 39). I'm figuring out a way around that.

alexgibson:
Chrome is an evergreen browser that auto updates (unlike IE). It looks like 39 was released in 2014, almost a decade ago. We probably don’t have to worry there.

We also feature detect matchMedia before using it in our code I believe, which should avoid throwing errors in very old browsers at least.

reemhamz:
@alexgibson Agreed, I thought I was overthinking it by testing out the evergreen browsers, but I just wanted to make sure the change landed well in our code. Thanks!

reemhamz:
I'm going to mark this as ready for review, but also ready for just general discussion in case we need to hold off that review, because:
I've been experiencing bug with Browserstack, specifically to do with the [sticky nav component](http://localhost:3000/components/detail/navigation--default). When testing both the localhost and prod site on Browserstack, the sticky nav is just not functioning, even when I choose to test it on the latest versions of Chrome and Firefox. It does however work in https://mozilla.org.

When I test Protocol's localhost and prod site on my browser (read: **not on Browserstack**), the sticky nav component works as expected.

I asked @craigcook about it, and he experienced a similar bug with Browserstack, but his Browserstack eventually resolved and no longer showed that bug, mine, on the other hand, did not resolve.

Curious if anyone else is experiencing the same bug.

alexgibson (approved):
One small nit but otherwise works great (tested in Firefox / IE) r+wc

Issue #847, opened 2023-02-02
JS should not add 'aria-expanded=false' to .mzp-c-navigation-items


stephaniehobson:
`aria-expanded=false` is being added to `.mzp-c-navigation-items`. 

This role should be declared on a controlling element like a button or link and must be used with `aria-controls`. Also, on desktop this content is expanded by default so it should not be false.

Wild speculation about the cause of this:

I think it should be on `c-navigation-menu-button` instead. Since the button is hidden I don't know if it would be confusing to a screen reader to have it declared false or not. We can fix it to be true on desktop or do research to see if that's necessary and document that it is not.

nathan-barrett:
Can this be closed? I think it was fixed in https://github.com/mozilla/protocol/pull/860

stephaniehobson:
Fixed in #860

PR #860, opened 2023-03-14
Update aria roles on mzp navigation

nathan-barrett:
## Description

Moves the `aria-expanded` attribute to the `mzp-c-navigation-menu-button` instead of the menu items as explained in #847. Also removes the aria-expanded attribute on desktop since the element is not displayed, so firefox no longer includes it in the accessibility tree.

### Issue

#847 

### Testing
You can test this by running a local server and navigating to this page: http://localhost:3000/components/preview/navigation--default and using the inspector to view the element and make sure that the aria-expanded role is toggling correctly.

### DIFF
--- a/assets/js/protocol/navigation.js
+++ b/assets/js/protocol/navigation.js
@@ -8,7 +8,7 @@ let _navItemsLists;
 let _navMenuButtons;
 const _options = {
     onNavOpen: null,
-    onNavClose: null,
+    onNavClose: null
 };
 let _ticking = false;
 let _lastKnownScrollPosition = 0;
@@ -185,26 +185,28 @@ MzpNavigation.onClick = (e) => {
  * use Intersection Observer API to keep track of when the mobile
  * nav menu is displayed to handle aria roles better
  */
-(MzpNavigation.menuButtonVisible = (callback) => {
+MzpNavigation.menuButtonVisible = (callback) => {
     // check if Intersection observer is supported
     if (MzpSupports !== 'undefined' && MzpSupports.intersectionObserver) {
         const observer = new IntersectionObserver(
             (entries) => {
-                entries.forEach((entry) => {
+                for (let index = 0; index < entries.length; index++) {
+                    const entry = entries[index];
                     callback(entry.intersectionRatio > 0, entry.target);
-                });
+                }
             },
             { root: document.documentElement }
         );
-        _navMenuButtons.forEach((button) => {
+        for (let index = 0; index < _navMenuButtons.length; index++) {
+            const button = _navMenuButtons[index];
             observer.observe(button);
-        });
+        }
     }
-}),
+},
 /**
-     * Set initial ARIA navigation states.
-     */
-(MzpNavigation.setAria = () => {
+ * Set initial ARIA navigation states.
+ */
+MzpNavigation.setAria = () => {
     if (MzpSupports !== 'undefined' && MzpSupports.intersectionObserver) {
         MzpNavigation.menuButtonVisible((isVisible, menuButton) => {
             if (isVisible) {
@@ -227,7 +229,7 @@ MzpNavigation.onClick = (e) => {
             menuButton.setAttribute('aria-expanded', isActive);
         }
     }
-});
+};
 
 /**
  * Bind navigation event handlers.
--- a/assets/js/protocol/navigation.js
+++ b/assets/js/protocol/navigation.js
@@ -5,9 +5,10 @@
 const MzpNavigation = {};
 let _navElem;
 let _navItemsLists;
+let _navMenuButtons;
 const _options = {
     onNavOpen: null,
-    onNavClose: null
+    onNavClose: null,
 };
 let _ticking = false;
 let _lastKnownScrollPosition = 0;
@@ -33,11 +34,13 @@ MzpNavigation.isLargeViewport = () => {
  */
 MzpNavigation.supportsSticky = () => {
     if (typeof MzpSupports !== 'undefined') {
-        return MzpSupports.matchMedia &&
-                MzpSupports.classList &&
-                MzpSupports.requestAnimationFrame &&
-                MzpSupports.cssFeatureQueries &&
-                CSS.supports('position', 'sticky');
+        return (
+            MzpSupports.matchMedia &&
+            MzpSupports.classList &&
+            MzpSupports.requestAnimationFrame &&
+            MzpSupports.cssFeatureQueries &&
+            CSS.supports('position', 'sticky')
+        );
     } else {
         return false;
     }
@@ -50,7 +53,9 @@ MzpNavigation.supportsSticky = () => {
  */
 MzpNavigation.onScroll = () => {
     if (!_ticking) {
-        _animationFrameID = window.requestAnimationFrame(MzpNavigation.checkScrollPosition);
+        _animationFrameID = window.requestAnimationFrame(
+            MzpNavigation.checkScrollPosition
+        );
         _ticking = true;
     }
 };
@@ -60,7 +65,9 @@ MzpNavigation.onScroll = () => {
  */
 MzpNavigation.createSticky = () => {
     _viewport.classList.add('mzp-has-sticky-navigation');
-    _animationFrameID = window.requestAnimationFrame(MzpNavigation.checkScrollPosition);
+    _animationFrameID = window.requestAnimationFrame(
+        MzpNavigation.checkScrollPosition
+    );
     window.addEventListener('scroll', MzpNavigation.onScroll, false);
 };
 
@@ -85,7 +92,9 @@ MzpNavigation.destroySticky = () => {
  * for sticky navigation are satisfied.
  */
 MzpNavigation.initSticky = () => {
-    _mqLargeNav = matchMedia(`(min-width: ${_wideBreakpoint}) and (min-height: ${_tallBreakpoint})`);
+    _mqLargeNav = matchMedia(
+        `(min-width: ${_wideBreakpoint}) and (min-height: ${_tallBreakpoint})`
+    );
 
     function makeStickyNav(mq) {
         if (mq.matches) {
@@ -143,7 +152,9 @@ MzpNavigation.checkScrollPosition = () => {
  * Event handler for navigation menu button `click` events.
  */
 MzpNavigation.onClick = (e) => {
-    const thisNavItemList = e.target.parentNode.querySelector('.mzp-c-navigation-items');
+    const thisNavItemList = e.target.parentNode.querySelector(
+        '.mzp-c-navigation-items'
+    );
 
     e.preventDefault();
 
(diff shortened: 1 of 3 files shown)


assets/js/protocol/protocol-navigation.js
```diff
@@ -169,14 +171,36 @@ if (typeof window.Mzp === 'undefined') { // eslint-disable-line block-scoped-var
             }
         }
     };
+    /**
+     * use Intersection Observer API to keep track of when the mobile
+     * nav menu is displayed to handle aria roles better
+     */
+
+    Navigation.menuButtonVisible = function (callback) {
+        var observer = new IntersectionObserver(function(entries) {
```
alexgibson, line 180: Might this throw an error in older browsers?

nathan-barrett: ack - you're right. Not supported in IE and browsers pre-2017ish - I'll need to look how to cover this in older browsers. Thanks for calling it out

alexgibson: It might be possible for you to use this and add a feature detect maybe, unless there are other options you think of

assets/js/protocol/protocol-navigation.js
```diff
@@ -185,16 +222,14 @@ if (typeof window.Mzp === 'undefined') { // eslint-disable-line block-scoped-var
     Navigation.bindEvents = function() {
         _navItemsLists = document.querySelectorAll('.mzp-c-navigation-items');
         if (_navItemsLists.length > 0) {
-            var navButtons = document.querySelectorAll('.mzp-c-navigation-menu-button');
-            for (var i = 0; i < navButtons.length; i++) {
-                navButtons[i].addEventListener('click', Navigation.onClick, false);
-            }
+            _navMenuButton = document.querySelector('.mzp-c-navigation-menu-button');
```
stephaniehobson, line 225: Switching from `querySelectorAll` to `querySelector` removes the ability to have more than one nav menu on the same page. We're not currently using this ability but it's a very common pattern, WMO could be using it on pages with a nav and sub nav. (This raises the question... why aren't we).

I think we should preserve that capability.

nathan-barrett: Thanks for bringing that up - I didn't think about that scenario. 👍🏻

assets/js/protocol/protocol-navigation.js
```diff
@@ -191,16 +231,16 @@ if (typeof window.Mzp === 'undefined') { // eslint-disable-line block-scoped-var
...
+            _navMenuButtons.forEach(function (menuButton) {
+                menuButton.addEventListener('click', Navigation.onClick, false);
+            });
             Navigation.setAria();
         }
     };
 
     /**
-     * Initialise menu.
+     * Initialize menu.
```
stephaniehobson, line 243: 🇺🇸

nathan-barrett: is it! oops - it just showed up on the spell checker 😿

stephaniehobson: Sorry, I just meant to tease for the change from 🇨🇦 to 🇺🇸 . It's a good change.

assets/js/protocol/protocol-navigation.js
```diff
@@ -175,13 +177,51 @@ if (typeof window.Mzp === 'undefined') { // eslint-disable-line block-scoped-var
...
+    /**
+     * use Intersection Observer API to keep track of when the mobile
+     * nav menu is displayed to handle aria roles better
+     */
+    Navigation.menuButtonVisible = function (callback) {
+        // check if Intersection observer is supported
+        if (!('IntersectionObserver' in window) ||
+            !('IntersectionObserverEntry' in window) ||
+            !('intersectionRatio' in window.IntersectionObserverEntry.prototype)) {
+            // load polyfill now
```
alexgibson, line 189: Question: Is there anything that should happen in this condition here? Adding a global polyfill for this feels like a bit too much perhaps. Are we just degrading gracefully instead, or should something else need to happen still? Maybe updating the comment here to that effect might be enough :)

assets/js/protocol/protocol-navigation.js
```diff
@@ -191,10 +231,10 @@ if (typeof window.Mzp === 'undefined') { // eslint-disable-line block-scoped-var
     Navigation.bindEvents = function() {
         _navItemsLists = document.querySelectorAll('.mzp-c-navigation-items');
         if (_navItemsLists.length > 0) {
-            var navButtons = document.querySelectorAll('.mzp-c-navigation-menu-button');
-            for (var i = 0; i < navButtons.length; i++) {
-                navButtons[i].addEventListener('click', Navigation.onClick, false);
-            }
+            _navMenuButtons = document.querySelectorAll('.mzp-c-navigation-menu-button');
+            _navMenuButtons.forEach(function (menuButton) {
```
alexgibson, line 235: Issue (blocking) This throws an error in IE: `Object doesn't support property of method 'forEach'`, since in older browsers a NodeList is treated differently to a regular array. Sticking to a conventional `for` loop might be better here.

alexgibson (changes requested):
Sorry this took me a while to get round to looking at. It looks good! I like the idea of using IntersectionObserver for things like this 👍 

A few small issues / fix-ups below, but it works well for me :)

assets/js/protocol/protocol-navigation.js
```diff
@@ -175,13 +177,51 @@ if (typeof window.Mzp === 'undefined') { // eslint-disable-line block-scoped-var
...
-            _navItemsLists[i].setAttribute('aria-expanded', false);
+        if (!('IntersectionObserver' in window) ||
+        !('IntersectionObserverEntry' in window) ||
+        !('intersectionRatio' in window.IntersectionObserverEntry.prototype)) {
+            _navMenuButtons.forEach(function (menuButton) {
+                var isActive = menuButton.classList.contains('mzp-is-active') && getComputedStyle(menuButton).display !== 'none';
+                menuButton.setAttribute('aria-expanded', isActive);
+            });
+        } else {
+            this.menuButtonVisible(function (isVisible, menuButton) {
```
alexgibson, line 214: Nit:
```suggestion
            Navigation.menuButtonVisible(function (isVisible, menuButton) {
```

assets/js/protocol/protocol-navigation.js
nathan-barrett: Oh shoot, didn't realize that - thank you

assets/js/protocol/protocol-navigation.js
```diff
@@ -175,13 +177,51 @@ if (typeof window.Mzp === 'undefined') { // eslint-disable-line block-scoped-var
...
 
     /**
      * Set initial ARIA navigation states.
      */
     Navigation.setAria = function() {
-        for (var i = 0; i < _navItemsLists.length; i++) {
-            _navItemsLists[i].setAttribute('aria-expanded', false);
+        if (!('IntersectionObserver' in window) ||
+        !('IntersectionObserverEntry' in window) ||
+        !('intersectionRatio' in window.IntersectionObserverEntry.prototype)) {
```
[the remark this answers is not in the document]
nathan-barrett: good idea 👍🏻

assets/js/protocol/protocol-navigation.js
nathan-barrett: I think just degrading gracefully - will update the coment

nathan-barrett:
Thanks @alexgibson - thought InteractionObserver was the best way to detect the change in CSS when an element is going from display:none to being displayed on a page.

assets/js/protocol/protocol-navigation.js
```diff
@@ -175,13 +177,46 @@ if (typeof window.Mzp === 'undefined') { // eslint-disable-line block-scoped-var
             }
         }
     };
+    /**
+     * use Intersection Observer API to keep track of when the mobile
+     * nav menu is displayed to handle aria roles better
+     */
+    Navigation.menuButtonVisible = function (callback) {
+        // check if Intersection observer is supported
+        if (Mzp.Supports !== 'undefined' && Mzp.Supports.intersectionRatio) {
```
alexgibson, line 186: Question: should this be `intersectionObserver`?
```suggestion
        if (Mzp.Supports !== 'undefined' && Mzp.Supports.intersectionObserver) {
```

assets/js/protocol/protocol-navigation.js
```diff
@@ -175,13 +177,46 @@ if (typeof window.Mzp === 'undefined') { // eslint-disable-line block-scoped-var
...
+        }
+    },
 
     /**
      * Set initial ARIA navigation states.
      */
     Navigation.setAria = function() {
-        for (var i = 0; i < _navItemsLists.length; i++) {
-            _navItemsLists[i].setAttribute('aria-expanded', false);
+        if (Mzp.Supports !== 'undefined' && Mzp.Supports.intersectionRatio) {
```
alexgibson, line 202:
```suggestion
        if (Mzp.Supports !== 'undefined' && Mzp.Supports.intersectionObserver) {
```

alexgibson (changes requested):
Suggested fixes look good, but I'm not sure if the new code is triggering? It seems to be falling back to the else clause (because the feature detect is named slightly differently I think).

Also, let's make sure we update the changelog for this bug fix before we merge! 👍

alexgibson:
There were some fairly big changes that landed in https://github.com/mozilla/protocol/pull/867, so to (hopefully) make things easier I took a go at rebasing your branch here. I opened a new PR, to make viewing the changes easier. I didn't change or update anything, so probably best check I didn't miss anything?

Feel free to use my commit / branch below in whatever way works best here:

https://github.com/mozilla/protocol/pull/873

nathan-barrett:
@alexgibson Thanks for making the PR! It was helpful as I was going through the rebase - I made sure that I was using the right MzpSupports check 🤦🏻 - let me know if you see anything that should be flagged

assets/js/protocol/navigation.js
```diff
@@ -5,9 +5,10 @@
 const MzpNavigation = {};
 let _navElem;
 let _navItemsLists;
+let _navMenuButtons;
 const _options = {
     onNavOpen: null,
-    onNavClose: null
+    onNavClose: null,
```
alexgibson, line 11: Nit: remove trailing comma (old browsers will throw an error on dangly's)

assets/js/protocol/navigation.js
```diff
@@ -169,23 +182,65 @@ MzpNavigation.onClick = (e) => {
...
+        );
+        _navMenuButtons.forEach((button) => {
+            observer.observe(button);
+        });
     }
-};
+}),
+/**
+     * Set initial ARIA navigation states.
+     */
```
alexgibson, line 206: Nit: fix comment intentation

assets/js/protocol/navigation.js
```diff
@@ -169,23 +182,65 @@ MzpNavigation.onClick = (e) => {
...
+(MzpNavigation.setAria = () => {
```
alexgibson, line 207: Question: Is there a reason why this function is wrapped in a parenthesis?

alexgibson (changes requested):
This needs a rebase, but aside from a couple of minor nits / suggestions the updates look good r+wc 👍

assets/js/protocol/navigation.js
```diff
@@ -169,23 +182,65 @@ MzpNavigation.onClick = (e) => {
...
+            (entries) => {
+                entries.forEach((entry) => {
+                    callback(entry.intersectionRatio > 0, entry.target);
+                });
+            },
+            { root: document.documentElement }
+        );
+        _navMenuButtons.forEach((button) => {
+            observer.observe(button);
+        });
```
alexgibson, line 201: Suggestion: there's still a `forEach` here on a node list here. It might be ok as most browsers that support Intersection Observer also seem to support `forEach`? But it also might be easier to stick to the same method throughout this file to avoid accidental future error?

assets/js/protocol/navigation.js
nathan-barrett: til!

PR #862, opened 2023-03-15
Fix syntax error in legacy IE when compiling JS (Fixes #861)

### LINKED ISSUE #861: Protocol components throw JS error in IE
## Description

Describe the issue that you're seeing.

### Steps to reproduce

1. In IE11, open https://protocol.mozilla.org/components/detail/modal.html
2. Open Dev Tools, and refresh the page to see the JS error.

### Expected result

No error

### Actual result

`SCRIPT1002 Syntax error` when opening `protocol-modal.js`.

Upon inspecting the JS file, the error seems to be because of an [arrow function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) inserted by Webpack when bundling.

The error is not present in bedrock, so it must be something specific to the Protocol config.

### Environment

IE11, Windows 10

alexgibson:
## Description

Adds missing browser list targets in `package.json` to make sure we're bundling JS in a compatible way with older IE browsers.

- [ ] ~I have documented this change in the design system~.

### Issue

#861

### Testing

I matched our browserlist targets in bedrock for parity: https://github.com/mozilla/bedrock/blob/main/package.json#L80-L83

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,5 +51,9 @@
     "karma-jasmine": "^5.1.0",
     "stylelint": "^14.16.1",
     "stylelint-config-standard-scss": "^6.1.0"
-  }
+  },
+  "browserslist": [
+    "defaults",
+    "IE 8"
+  ]
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# HEAD
+
+## Bug Fixes
+* **js:** Fix syntax error in legacy IE when compiling JS (#861)
+
 # 16.1.0
 
 ## Features


package.json
```diff
@@ -51,5 +51,9 @@
     "karma-jasmine": "^5.1.0",
     "stylelint": "^14.16.1",
     "stylelint-config-standard-scss": "^6.1.0"
-  }
+  },
+  "browserslist": [
+    "defaults",
+    "IE 8"
+  ]
```
nathan-barrett, line 58: Ah good catch @alexgibson!

nathan-barrett (approved):
r+!

Issue #594, opened 2020-06-01
Update Button Focus State to match other focus states within protocol


dzingeek:
After updating focus states for [text links](https://github.com/mozilla/protocol/issues/589) and [inline links](https://github.com/mozilla/protocol/issues/428) to make them more accessible we realized we should do the same for the buttons.

Rudresh22:
Hello, I want to work on this issue. Please guide me through this.

dzingeek:
@Rudresh22 here are the new specs for the focus state for our buttons. Please let me know if you have any questions or need more guidance. 

**image showing button focus state for reference**
![image showing button focus state for reference](https://user-images.githubusercontent.com/3740746/86183874-d01d1500-bae7-11ea-8602-8b8cc4928dae.png)


**focus state specs are as followed:**

- Border-radius: **2px for sm buttons and 4px for md, lg and lx buttons sizes**
- Outer border: **#0060DF 30%, 2px**  
- Middle border: **#FFFFFF, 1px**
-  Inner boarder: **#0060DF, 2px**

Rudresh22:
@dzingeek On it thanks.

Rudresh22:
Hello, @dzingeek ma'am, I seem to be having a linebreak error while running 'npm start' and hence I am not able to start the development server. Can I make the changes without running the development server?

dzingeek:
@craigcook can you help with the above?

craigcook:
What is the error?

`npm start` will install all dependencies, then build Protocol and start the server. Alternatively you can run `npm install` just to get the dependencies, then `gulp --dev` to build and start the server. Running them separately might help you isolate an error. And once it's installed you don't need to reinstall every time you want to work on it, you can just do `gulp --dev` from then on.

Rudresh22:
@dzingeek ma'am, I should make the changes for the given specs in the _button.scss file right? Very sorry for the inconvenience, I am a little new to OS.

craigcook:
Yes, the file is `components/_button.scss`

reemhamz:
Closed w/ #864

PR #864, opened 2023-03-23
Update focus state for buttons

reemhamz:
## Description
Updated our buttons' focus states to match what was listed the [issue's comment](https://github.com/mozilla/protocol/issues/594#issuecomment-652083403).

### Issue

#594 

### Testing

Test out all the button variants: 
http://localhost:3000/components/detail/button--primary

### DIFF
--- a/assets/sass/protocol/components/_button.scss
+++ b/assets/sass/protocol/components/_button.scss
@@ -7,8 +7,11 @@
 
 @mixin default-states {
     &:focus {
-        border-color: $button-border-color-focus;
-        box-shadow: $field-focus-ring;
+        // border-color: $button-border-color-focus;
+        // box-shadow: $field-focus-ring;
+        border: 2px solid #008AEA;
+        outline: $field-focus-ring-dark;
+        outline-offset: 1px;
     }
 
     &:hover {
@@ -27,8 +30,12 @@
 
 @mixin default-states-dark {
     &:focus {
-        border-color: $button-border-color-focus;
-        box-shadow: $field-focus-ring-dark;
+        // border-color: $button-border-color-focus;
+        // box-shadow: $field-focus-ring-dark;
+        // outline-offset: 2px;
+        border: 2px solid #008AEA;
+        outline: $field-focus-ring-dark;
+        outline-offset: 1px;
     }
 
     &:hover {
@@ -47,8 +54,11 @@
 
 @mixin product-states {
     &:focus {
-        border-color: $button-border-color-focus;
-        box-shadow: $field-focus-ring;
+        // border-color: $button-border-color-focus;
+        // box-shadow: $field-focus-ring;
+        border: 2px solid #008AEA;
+        outline: $field-focus-ring-dark;
+        outline-offset: 1px;
     }
 
     &:hover {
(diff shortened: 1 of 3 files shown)


assets/sass/protocol/includes/forms/_lib.scss
```diff
@@ -29,11 +29,13 @@ $field-padding: $spacing-sm;
 $field-h-spacing: $spacing-md;
 $field-v-spacing: $spacing-lg;
 $field-focus-ring: 0 0 0 2px rgba($color-blue-40, 0.5);
-$field-focus-ring-dark: 0 0 0 2px color.mix($color-blue-40, $color-white);
+$field-focus-ring-dark: 0 0 0 2px $color-white;
+$field-focus-ring-outline: 2px solid #afccf3;
```
reemhamz, line 33: I added `#afccf3` instead of using `rgba($color-blue-50, 0.3)` because the set opacity was basically invisible in dark theme, so I got the hexcode of the lighter color and placed it in the code instead. Happy to hear of other suggestions.

assets/sass/protocol/includes/forms/_lib.scss
```diff
@@ -30,10 +30,13 @@ $field-h-spacing: $spacing-md;
 $field-v-spacing: $spacing-lg;
 $field-focus-ring: 0 0 0 2px rgba($color-blue-40, 0.5);
 $field-focus-ring-dark: 0 0 0 2px color.mix($color-blue-40, $color-white);
+$field-focus-ring-white: 0 0 0 2px $color-white;
+$field-focus-ring-outline: 2px solid #afccf3;
```
reemhamz, line 34: I added #afccf3 instead of using rgba($color-blue-50, 0.3) because the set opacity was basically invisible in dark theme, so I got the hexcode of the lighter color and placed it in the code instead. Happy to hear of other suggestions.

assets/sass/protocol/components/_button.scss
```diff
@@ -8,7 +8,8 @@
 @mixin default-states {
     &:focus {
         border-color: $button-border-color-focus;
```
maureenlholland, line 10: **nit (non-blocking, pre-existing)**: we have redundant border-color rules on neutral buttons

**suggestion**: while we're here, let's remove the redundant rules under `mzp-t-neutral` styles. Neutral buttons already inherit default state styles from `mzp-c-button`. We should only override the border-color property if it needs to be different.
neutral

maureenlholland (approved):
LGTM 👍 

nothing blocking, very minor suggestions for hex code comment and removal of redundant style rule

assets/sass/protocol/includes/forms/_lib.scss
reemhamz: This actually was a pretty good idea. I played around with Sass' color function but I hadn't thought of using `saturation` here 👍🏼

assets/sass/protocol/components/_button.scss
reemhamz: The neutral states have their own state styles different from the primary and secondary buttons, so if I were to remove their styles, we'd be removing their state styles entirely. The other button styles are able to inherit default styles because the default/product styles mixins are being passed into the respective classes using the `@include` rule. 

Also, if I do remove the neutral states and pass along the default-styles mixin, I'd essentially be getting rid of  the neutral styles we have embedded. 🤕

maureenlholland: Sorry, I think my suggestion was badly worded. The only redundant rules I would want to remove is the Neutral button :focus border-color on L222 & L244. They're not actually using a custom "neutral" focus color, but sharing the `$button-border-color-focus` variable.

The more I look at it, actually, the more I find it strange we re-declare the `default-states` in `mzp-t-button-secondary` too
redundant
inherited

I'm going to create a separate issue to address this. Let's merge these focus states!

maureenlholland: In the end, I couldn't remove the secondary rule because it would require use of `:not(:hover, :focus, :active)` selector on the theme styles, which isn't supported in IE11. Leaving this in draft form: https://github.com/mozilla/protocol/pull/866
You can incorporate it here or merge as-is and I'll update the changelog on the draft PR

maureenlholland (approved):
r++ 🟢

assets/sass/protocol/components/_button.scss
reemhamz: hahaha, thanks!

PR #866, opened 2023-03-27
Remove redundant styles (fixes #865)

### LINKED ISSUE #865: Remove redundant button component style rules
We have a few redundant state style rules in our themed buttons. Secondary & Neutral buttons should only override default styles, not repeat them.
redundant
inherited

maureenlholland:
## Description

### Issue

https://github.com/mozilla/protocol/issues/865

### Testing
Focus state for neutral button should be the same in local
http://localhost:3000/components/detail/button--neutral
& prod
https://protocol.mozilla.org/components/detail/button--neutral.html

### DIFF
--- a/assets/sass/protocol/components/_button.scss
+++ b/assets/sass/protocol/components/_button.scss
@@ -142,6 +142,7 @@ a.mzp-c-button {
     border: 2px solid $color-black;
     color: $color-white;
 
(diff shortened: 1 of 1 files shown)


maureenlholland:
The repetition is necessary for higher specificity in both cases. It's confusing to look at, but it's there for a reason.

PR #867, opened 2023-04-04
Convert Protocol JS components to modern JS formats (Fixes #255)

### LINKED ISSUE #255: Convert Protocol ES5 to modern JS
Since this issue was originally opened Bedrock has migrated much of its JS from ES5 to modern JS syntax, and now supports things like ES modules. We use webpack to transpile to ES5 for backward compatibility.

I'd like to do the same here for Protocol's JS components, so that we can enjoy more parity between writing for Protocol and writing for Bedrock.

Doing this will make it easier for other websites to consume Protocol in a modern build system. It will also unlock backporting things from Bedrock to Protocol, such as the Basket newsletter logic.

The Protocol JS components should be output in UMD format, to support the widest possible range of import / usage methods.

We'll need to test to make sure things import into Bedrock OK as we go.

alexgibson:
## Description

This PR updates all Protocol JS components so that they can be authored using modern JavaScript syntax and features. We use Webpack to transpile the components back to ES5 and [UMD format](https://github.com/umdjs/umd) for publishing.

This PR also removes pre-minified JS and CSS files from the NPM package.

This is a large-ish, breaking change, so requires some careful review.

### Issue

#255

### Testing

A couple of useful points when testing this PR:

- I simplified some of our NPM scripts when building the site. When running locally it's now just `npm run webpack` instead of `npm run webpack-docs`.
- We no longer have a global `window.Mzp` object! Components have their own global object now, so things like `window.Mzp.Modal` become `window.MzpModal`.

Demo: https://demo1--mozilla-protocol.netlify.app/

I also have a draft bedrock PR here that tests the changes locally: https://github.com/mozilla/bedrock/pull/12981


alexgibson:
Ok, I think this is finally ready for review :)

alexgibson:
Note for reviewer: the JS changes here look bigger than they are. I mostly just changed things like `var` to `const` or `let`, and used arrow functions where appropriate etc. The actual logic for the components remains unchanged.

components/details-component/readme.md
```diff
@@ -3,10 +3,46 @@ each heading into a control to expand or collapse all the content that comes aft
...
+### Dependencies
+
+The Details component depends on both the `MzpSupports` and `MzpUtils` libraries for feature detection
+and DOM traversal. It is recommended that both are included in your page and accessible via a global
+object before loading your Details component script.
+
+```
+import MzpSupports from '@mozilla-protocol/core/protocol/js/supports');
+import MzpUtils from '@mozilla-protocol/core/protocol/js/utils');
+
```
nathan-barrett, line 38: Seem like there are extra `)` in these two lines

nathan-barrett (approved):
Awesome work Alex, a lot of changes but everything looks great after testing it out locally. Thanks for putting this together!

alexgibson:
@nathan-barrett thanks - rebased and updated

alexgibson:
@craigcook this has been reviewed and approved, but I'll leave to you for merge / launch timings.

This is tied to one of my H1 goals, so as long as I can still get this back into bedrock before the end of the half then it should all be good 👍

PR #894, opened 2023-08-15
Fix incorrect rem value

nathan-barrett:
## Description
Protocol v18 had an incorrect value for `test-body-lg` variable in the firefox theme.

- [ x] I have documented this change in the design system.
- [ x] I have recorded this change in `CHANGELOG.md`.

### Testing
`text-body-lg` should be 1.125rem instead of 1.1rem


craigcook (approved):
👕

PR #905, opened 2023-10-26
Port basket newsletter JS to Protocol (Fixes #839)

### LINKED ISSUE #839: Update newsletter component to include JS to post directly to Basket

Now that https://github.com/mozilla/bedrock/issues/11944 is complete, newsletter forms no longer rely on bedrocks' backend to post to Basket. There are still a few external sites that post via bedrock however, so we'd like to provide a go-to alternative here in Protocol that they can use / cherry-pick instead.

The [JS code in bedrock](https://github.com/mozilla/bedrock/blob/main/media/js/newsletter/newsletter.es6.js) should be moved into Protocol so that other sites can more easily use it.

### Please describe why these changes need to be made

So that Mozilla sites can more easily post to basket without relying on bedrock.

### Please provide any designs or prototypes of the proposed changes

N/A

All Protocol newsletter component forms on bedrock currently use this code.

N/A

We'd like to remove the backend functionality from bedrock entirely, so making this happen would be helpful in providing an alternative.

alexgibson:
## Description

This PR ports the JS code (and tests) from bedrock that handles newsletter form submissions.

This will help provide Mozilla websites with an up to date example of how to handle newsletter submissions that post directly to Basket, and will act as a replacement for https://github.com/mozilla/basket-example.

I also fixed a couple of other pesky long time bugs, like how small `<legend>` elements were rendered in Zilla Slab, making them hard to read.

### Issue

#839

### Testing

- http://localhost:3000/components/detail/newsletter--default
- http://localhost:3000/components/detail/newsletter--errors
- http://localhost:3000/components/detail/newsletter--success

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,7 +3,6 @@ module.exports = {
         'browser': true,
         'commonjs': true,
         'es2017': true,
-        'jasmine': true,
         'node': true
     },
     extends: [
@@ -74,5 +73,17 @@ module.exports = {
         // Disallow the use of `console`
         // https://eslint.org/docs/rules/no-console
         'no-console': 'error'
-    }
+    },
+    overrides: [
+        {
+            // JS Jasmine test files.
+            files: ['tests/unit/**/*.js'],
+            env: {
+                jasmine: true
+            },
+            globals: {
+                sinon: 'writable'
+            }
+        }
+    ]
 };
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Features
 
+* **js:** Update newsletter component to include JS to post directly to Basket (#839).
 * **css:** Add CSS utility class for centered text and document existing title utility classes (#897).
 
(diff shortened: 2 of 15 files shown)


reemhamz (approved):
It works!
I tested it out, and I saw the form submission go to Basket in dev tools.

However, instead of a `POST` request, I'm getting an `OPTIONS` request. Is this expected? I'm assuming that's the case because it's determining if the URL it wants to `POST` to is safe to send?

In any case, r+(wc?) 🍅

alexgibson:
@reemhamz thanks for the review!

The way CORS (cross origin) requests work is that browsers will first send an `OPTIONS` pre-flight request to make sure the site is allowed to make the cross origin request. If allowed, it will then make the `POST`. You should hopefully see two requests in the network console. 

image

reemhamz:
Ah yup, I see both requests in the console :) so this is good to merge, but I'll leave that to you! 👍🏼

alexgibson:
Thanks for double checking!

PR #908, opened 2023-12-07
Add refactored Callout component, replace Call-out [fix #787]

### LINKED ISSUE #787: Deprecate (or update) the Callout component
We've deprecated the Hero component in favor of Split (for Hero with image) or Callout (for Hero without an image). However, Callout in its current state isn't actually an adequate replacement for the non-image, centered Hero without some modifications.

Namely:
- Callout has built-in logos. We should instead prefer the newer Logo/Wordmark components (however we first need to make Logos and Wordmarks centerable, see #718 which would be a blocker for this).
- Hero has a max-width of 580px on its content; Callout contains an `mzp-l-content` container with a max-width of 800px, which breaks from the standard container widths. Callout should support the standard container widths.
- The Callout title has a set size, a bit smaller than the Hero title, since it wasn't designed to be a page title. Instead of a set size for the title we could instead use utility classes for title sizes.
- Callout has a default background color. It should be transparent by default, with a background added by a theme class or local styling.

We could make all these changes to the Callout component **OR** we could create a new component with the properties we want and deprecate Callout entirely. 

I'm leaning toward deprecation mainly because some of these would be breaking changes, especially removing the built-in logos. But if we make a new component and add a deprecation warning for Callout that would allow people more time to transition.

A new, content-agnostic, centered section with some spacing, standardized content widths, and an optional background color could replace a centered Hero as well as anywhere else we currently use Callout.

craigcook:
## Description

This refactors the Callout, but also changes the name so it effectively removes and replaces the previous component.

The revised Callout:
1. Removes integrated branding options. We can use Logo/Wordmark components for branding.
2. Treats Compact Callout as a layout variant rather than as a separate component.

This also means a Callout can replace a default centered Hero, and we can finally fully remove the Hero component.

However, since this renames/replaces a component it's a breaking change and will need a major release. We could keep the old name but I prefer it without a dash. But moreover I wanted to make the compact version a variant and not a whole component, so that would deprecate Compact Call-out anyway. 

So... if we really care about gradual deprecation and keeping old components for a while to let people transition, we shouldn't change the name. But I want to change the name, and keeping the dashed and undashed names is even messier than just doing it.

### Issue
#787 

### Testing

https://demo2--mozilla-protocol.netlify.app/components/detail/callout--default
http://localhost:3000/components/detail/callout--default


components/hero/readme.md
```diff
@@ -1,6 +1,6 @@
-**This component is deprecated.** You can use the [Call-out component](call-out)
-in place of a centered Hero, or the [Split component](split)
-in place of a Hero with image.
+**This component is deprecated.** You can use the [Display component](display)
```
nathan-barrett, line 1: Think this should be changed to something else?

nathan-barrett (approved):
rwc+ Thats a great looking callout. Just some minor documentation questions

components/hero/readme.md
craigcook: Doh, good catch... that's a leftover from when I couldn't think of a name.

PR #916, opened 2024-01-09
Center-align button text [fix #826]

### LINKED ISSUE #826: Button text should be centered
## Description

When button styling is applied to a link (`a` element) it inherits `text-align` from its parent. `button` and `input type="button"` have centered text as part of browser default styling but links don't. 

We should explicitly add `text-align: center` to our base button style to normalize across both `button`.

craigcook:
## Description

Explicitly sets `text-align: center;` for `.mzp-c-button` to compensate for button-styled links inheriting their alignment from their parent. This only really manifests when a button-style link is set to the full width of its container, but that does happen sometimes.

### Issue
#826


reemhamz (approved):
🏯

PR #925, opened 2024-03-25
Add optional icon to buttons [fix #893]

### LINKED ISSUE #893: Update buttons to accept an optional icon
Sometimes we want to put an icon on a button, to make it distinct from other buttons or indicate a specific function. For example, a download button might include a little down arrow icon like: image

A few approaches we might take:

1. Update button styling to accommodate an inline image, so we could include icons arbitrarily (maybe wrapped in a span, might need a modifier class on the button).
2. Predefine a few specific kinds of icon buttons for specific purposes, then a theme class could handle the icon in CSS (a pseudo-element with a background).

Additional thoughts:

* We might need style alternatives for icon-on-left or icon-on-right, unless we decide to commit to only one.
* Icon position should respect language direction.
* We probably shouldn't allow doubling icons (no left AND right). Unless there's a very good use case for two icons in a button?
* If we use Protocol icon assets we'll need them in both dark and light (though if it's an inline SVG we could color the icon with CSS)

craigcook:
## Description
Allows an inline icon in buttons.

### Issue
#893 

### Testing
https://demo2--mozilla-protocol.netlify.app/components/detail/button--with-icon
http://localhost:3000/components/detail/button--with-icon

### DIFF
--- a/components/button/button--with-icon.html
+++ b/components/button/button--with-icon.html
@@ -1,0 +1,63 @@
+<p>
+{% render '@button', { label: 'Button with Icon',
+  iconEnd:
+  '<svg xmlns="http://www.w3.org/2000/svg" xml:space="preserve" viewBox="0 0 24 24" width="24" height="24" fill="currentColor">
+    <path d="m21.3 4.5-8-4c-.8-.4-1.8-.4-2.7 0l-8 4C1.6 5.1 1 6.1 1 7.2v9.5c0 1.1.6 2.2 1.7 2.7l8 4c.4.2.9.3 1.3.3s.9-.1 1.3-.3l8-4c1-.5 1.7-1.5 1.7-2.7V7.2c0-1.1-.6-2.1-1.7-2.7zm-9.7-2.2c.1-.1.3-.1.4-.1.2 0 .3 0 .4.1L19.8 6 12 9.9 4.2 6l7.4-3.7zm-8 15.4c-.4-.2-.6-.6-.6-.9V7.6l8 4v9.8l-7.4-3.7zm16.8 0L13 21.4v-9.8l8-4v9.2c0 .3-.2.7-.6.9z"/>
+  </svg>' }
+%}
+</p>
+
+<p>
+{% render '@button', { label: 'Button with Icon',
+  iconStart:
+  '<svg xmlns="http://www.w3.org/2000/svg" xml:space="preserve" viewBox="0 0 24 24" width="24" height="24" fill="currentColor">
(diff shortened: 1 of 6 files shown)


stephaniehobson:
What are your thoughts on using `:first-child` and `::last-child` to add the icon spacing instead of a class on the parent?

stephaniehobson (approved):
Hover works like a charm. Nice use of `em` and `ch`.

R+ ➡️  unless you want to totally re-write it to avoid the class on the parent.

craigcook:
> What are your thoughts on using `:first-child` and `::last-child` to add the icon spacing instead of a class on the parent?

Oooo there's an idea... I'll give it a try.

craigcook:
To follow up on the above comments, for posterity...

Using `:first-child` and `:last-child` ended up not working because the button label is only text, and the icon ends up as the ONLY child. We could wrap the label in another span but that seemed unnecessary. BUT we can still avoid the need for a class on the parent button by applying the -start and -end styling on the icon itself (it's really only needed for spacing).

So now the icon itself takes the class `mzp-c-button-icon-start` or `mzp-c-button-icon-end` and no class is needed on the parent button.

Also I tried applying these classes directly to the icon image without wrapping it in a span, but the span helps normalize the size and position so it still works best with a span. Then the contents of that span could be an SVG, an `img` or even a text character or emoji; the span is content-agnostic.

components/button/button--with-icon.html
```diff
@@ -0,0 +1,63 @@
...
+{% render '@button', { label: 'Button with Icon',
+  iconStart:
+  '<svg xmlns="http://www.w3.org/2000/svg" xml:space="preserve" viewBox="0 0 24 24" width="24" height="24" fill="currentColor">
+    <path d="m21.3 4.5-8-4c-.8-.4-1.8-.4-2.7 0l-8 4C1.6 5.1 1 6.1 1 7.2v9.5c0 1.1.6 2.2 1.7 2.7l8 4c.4.2.9.3 1.3.3s.9-.1 1.3-.3l8-4c1-.5 1.7-1.5 1.7-2.7V7.2c0-1.1-.6-2.1-1.7-2.7zm-9.7-2.2c.1-.1.3-.1.4-.1.2 0 .3 0 .4.1L19.8 6 12 9.9 4.2 6l7.4-3.7zm-8 15.4c-.4-.2-.6-.6-.6-.9V7.6l8 4v9.8l-7.4-3.7zm16.8 0L13 21.4v-9.8l8-4v9.2c0 .3-.2.7-.6.9z"/>
+  </svg>' }
+%}
+</p>
+
+<p>
+{% render '@button', { class: 'mzp-t-xl mzp-t-product', label: 'Download Firefox', link: 'https://www.mozilla.org/firefox/new/',
```
reemhamz, line 20: **Suggestion:** I still believe it would be worthwhile to create a class for the download button so it can be easily updated throughout Bedrock

reemhamz (approved):
You did a pretty clean job with this update 🧼 
I left a suggestion that I strongly think we should implement.

I'll leave this as r+wc because I'm sure it's a simple change from your end if you choose to go w/ the suggestion

components/button/button--with-icon.html
craigcook: I think we may end up designing a download button as a specific component down the road, and that could have an icon built in. I'll file a separate issue for that.

Issue #930, opened 2024-04-22
Firefox Klar [DE] logo classes missing


janbrasna:
## Description

There's no logo / wordmark components for Firefox Klar (Firefox Focus in German) generated for `mzp-t-product-klar` class as one would expect, although the product assets are otherwise available.

The product's not included anywhere branding's a topic e.g. [logo--products.readme.md](https://github.com/mozilla/protocol/blob/294d66697683c4dc6be5a810cc6733fa73d87f4e/components/branding/logo/logo--products.readme.md?plain=1#L7)

### Steps to reproduce

`<div class="mzp-c-wordmark mzp-t-wordmark-md mzp-t-product-klar">Firefox Klar</div>`

### Expected result

`mzp-t-product-klar` would produce the correct brand.

### Actual result

None. For now `mzp-t-product-focus` has to be used, and/or the correct asset override specified manually.

PR #932, opened 2024-04-30
Add Firefox Klar wordmark

reemhamz:
## Description
We do not have a CSS class defined for Firefox Klar wordmark to be used for German locales in Bedrock. Instead, we've been using the English Firefox Focus wordmark. This PR adds that classname so it can be used (with a `{% if LANG=='de' %}` conditional.

i.e. (example used from @janbrasna's [PR in Bedrock](https://github.com/mozilla/bedrock/pull/14490/files#diff-f97847d5549ff9ef28dc0eabf97fe4f7b7a13d41a355fa1c7d8537c967d7604aR60):

```
<div class="mzp-c-wordmark mzp-t-wordmark-md {% if LANG=='de' %} mzp-t-product-klar {% else %}mzp-t-product-focus{% endif %}">{{ ftl('mobile-focus-firefox-focus') }}</div>
```

>[!NOTE]
> I didn't need to add a new logo component to support Klar since it's just Firefox Focus in German, so it uses the same logo.

## Describe what this change does.

- [x] Imports the Firefox Klar wordmark
- [x] Creates a class for Firefox Klar to be used in Bedrock (`mzp-t-product-klar` in conjunction with `mzp-c-wordmark`)

## Issue

Fixes https://github.com/mozilla/protocol/issues/930
Refs https://github.com/mozilla/bedrock/pull/14490

## Testing

http://localhost:3000/components/detail/wordmark--default -- change the `firefox` string to `klar` and you should see the wordmark appear on the page

### DIFF
--- a/assets/sass/protocol/components/logos/_wordmark-product-klar.scss
+++ b/assets/sass/protocol/components/logos/_wordmark-product-klar.scss
@@ -1,0 +1,12 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@use 'wordmark';
+
+$class: 'klar';
+$dir: 'firefox/browser/klar';
+
+@each $layout-size, $logo-size in wordmark.$logo-sizes {
+    @include wordmark.wordmark($class, $dir, $layout-size, $logo-size);
+}
--- a/components/branding/wordmark/wordmark--products.readme.md
+++ b/components/branding/wordmark/wordmark--products.readme.md
@@ -5,6 +5,7 @@ Product classes:
 - `mzp-t-product-developer`
 - `mzp-t-product-nightly`
 - `mzp-t-product-focus`
+- `mzp-t-product-klar`
 - `mzp-t-product-lockwise`
 - `mzp-t-product-monitor`
 - `mzp-t-product-mozilla`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,13 @@
+# HEAD
+
+## Features
+* **component:** Add Firefox Klar wordmark for the Wordmark component
+
 # 19.1.0
 
 ## Features
 
-* **component** Remove the format option from Newsletter component (#926)
+* **component:** Remove the format option from Newsletter component (#926)
 * **component:** Allow an optional icon in buttons (#893)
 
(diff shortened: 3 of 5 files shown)


stephaniehobson (approved):
R+

assets/sass/protocol/components/logos/_wordmark-product-klar.scss
```diff
@@ -0,0 +1,12 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@use 'wordmark';
+
+$class: 'klar';
+$dir: 'firefox/browser/klar';
+
+@each $layout-size, $logo-size in wordmark.$logo-sizes {
+    @include wordmark.wordmark($class, $dir, $layout-size, $logo-size);
```
stephaniehobson, line 11: **Thought:** Is there any benefit to including the klar classes in the focus file? Are we likely to use one without the other?

reemhamz: Benefit? Probably, but I didn't want to mess up how the CSS variables are structured for wordmarks! I'm also not sure I know how to set up conditional logic to extract the Klar workmark only in German only through this file. Because we import one wordmark per file, we'd make to use more specification in how we import this into Bedrock's CSS file, too.

I'm happy to merge for now. Let me know if you think there's a way to cleanly do the above and we can implement that in another PR

components/branding/wordmark/wordmark--products.readme.md
```diff
@@ -5,6 +5,7 @@ Product classes:
 - `mzp-t-product-developer`
 - `mzp-t-product-nightly`
 - `mzp-t-product-focus`
+- `mzp-t-product-klar`
```
janbrasna, line 8: FYI there's the same list also in `wordmark/readme.md` that could benefit from this addition for consistency, if you think it's worth it.

I'm actually thinking about how to document properly that the Klar component is only among the wordmarks, and trying to use the product class with another branding element will fail and folks have to remember to use Focus classes for all other branding needs — so this is an odd duck, existing only for the text difference here, and not elsewhere where it has no use. Some kind of a note to warn people of how specific it is?

Xcopz1 (commented):
> image

Xcopz1 (commented):
[Untitled.pdf](https://github.com/mozilla/protocol/files/15213998/Untitled.pdf)

PR #949, opened 2024-07-03
Add `text-wrap: balance` to all headings

### LINKED ISSUE #910: Add text-wrap: balance to headings
The CSS property `text-wrap: balance;` is now supported in Firefox 121, and has pretty good support across browsers: https://developer.mozilla.org/en-US/docs/Web/CSS/text-wrap

I propose we should add this to all heading elements in Protocol's base styles as a general typographical improvement.

janbrasna:
## Description

> css-ui_css-text-wrap-balance

https://caniuse.com/css-text-wrap-balance is pretty well supported now with no real downside, known bugs or performance issues documented.

### Issue

Resolves #910

### Testing

Safari 16 vs. 17.5+
Firefox 115 (ESR) vs. 122+

PR: https://preview--mzp-dev.netlify.app/components/detail/headings

_(before:)_
Screenshot 2025-03-28 at 2 32 45 Large

_(after:)_
Screenshot 2025-03-28 at 2 32 59 Large

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
 # HEAD
 
 ## Features
-* **assets:** Upgrade Inter font to version 4.1, drop support for Inter `woff` (bedrock #15999)
 
+* **fonts:** Upgrade Inter font to version 4.1, drop WOFF 1.0 format (#1026)
+* **css:** Add `text-wrap: balance` to all headings (#910)
 * **css:** Apply hover cursor from Details component to Details element (#948)
 
 ## Bug Fixes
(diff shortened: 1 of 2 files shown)


stephaniehobson:
@janbrasna Do you remember what the problem is with `white-space`?

janbrasna:
@stephaniehobson So the issue is what risk this might pose for downstream consumers, where balancing text IIUC actually competes with `white-space` property per se (as one is asking for no wrapping and the other is asking for balanced wrapping) — guess I wanted to check a few layouts over at kitsune, fxa, relay et al. how such a change wholesale could interfere with what they have added to the cascade in addition to protocol styling.

(For just bedrock this should be a non-issue, as balancing headings has been around for some time now.)

stephaniehobson:
Seems like it follows the cascade in Firefox and Chrome so any custom styles they have applied should take president if they're importing their styles after the Protocol one.

I think we're safe to go ahead but I'd add something to the CHANGELOG's migration notes to search for any place where `white-space: nowrap` is defined and make sure it has a higher specificity.

stephaniehobson:
It might be a nice change to include with V20 which will have other font updates.

janbrasna:
@stephaniehobson Yup, on it. I'll try to add some reasonable heads up to the migration notes too.

(I haven't checked this is not also included in the wholesale font update in the other PR, porting other related values battle tested in bedrock, but will do so! — I actually found this to be a solution to an "unrelated" whitespace / hyphens / wordwrapping issue where having balanced headings helps avoiding a regression when allowing implicit hyphens, so can't wait to have this in and start removing weird forced word wrapping.)

Landing this in a major / breaking release makes the most sense; I was mainly cautious about slipping this in a minor update regarding any impact to consumers, but v20 is great to have this included in, being full of similar changes. 🚀

stephaniehobson (approved):
R+ ⚖️

PR #956, opened 2024-07-18
Fix modal "X" close cursor affordance

### LINKED ISSUE #955: Modal close button pointer overridden by defaults
## Description

Even though `.mzp-c-modal-close` has cursor: pointer defined, the actual child button `.mzp-c-modal-button-close` itself is not setting it and since it's a ‹button› element it gets a good deal of browser styling defaults applied to it, overriding the pointer.

### Steps to reproduce

See https://protocol.mozilla.org/components/detail/modal or visit e.g. https://www.mozilla.org/firefox/all/ where the ‹?› help popus are _modal_ instances.

### Expected result

The "X" close button should have pointer cursor to reconfirm it's an active element.

### Actual result

The parent has it defined, but the actual button is brutally overridden by UA defaults for button element, i.e. effectively no pointer.

### Environment

macOS Catalina
FXDE 129b5, Chromium 127r1313161

janbrasna:
## Description

Sets the cursor on the actual button element to avoid any UA defaults overriding it.

- [ ] ~I have documented this change in the design system.~

### Issue

Resolves #955

### Testing

https://fix-modal-close-pointer--mzp-dev.netlify.app/components/detail/modal

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## Features
 * **js:** Migrate tests from Karma to Jasmine browser runner (#899)
 
+## Bug Fixes
+* **css:** Modal close button pointer overridden by defaults (#955)
+
 # 19.2.0
 
 ## Features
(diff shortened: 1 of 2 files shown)


stephaniehobson (approved):
R+ 🙅‍♀️

PR #1026, opened 2025-03-07
update Inter font to version 4.1 (fix bedrock #15999)

wen-2018:
## Description

This PR updates the Inter font to version 4.1

### Issue

https://github.com/mozilla/bedrock/issues/15999

### Testing


maureenlholland:
Some previous exploration of variable fonts was done in 2022: https://github.com/mozilla/protocol/pull/810

We might want to look at if 3 or more Inter fonts are loaded on the new refresh designed pages to get a more current performance comparision

maureenlholland (commented):
LGTM r+wc ✍️ 

**suggestion (non-blocking)** update [CHANGELOG](https://github.com/mozilla/protocol/blob/main/CHANGELOG.md#head) to include the info about font version upgrade

Screenshot 2025-03-11 at 2 22 26 PM



seems a bit odd to me that the prod file sizes are smaller than the PR, but it's not a huge difference

PROD
prod

PR
pr

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # HEAD
 
+## Features
+* **assets:** Upgrade Inter font to version 4.1 (bedrock #15999)
+
+# 19.3.0
```
maureenlholland, line 6:
```suggestion
```

We haven't published since the last update, so you can place this under the existing HEAD Features with karma to jasmine update

maureenlholland (changes requested):
minor requested change on the changelog, then good to go

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Features
 * **js:** Migrate tests from Karma to Jasmine browser runner (#899)
+* **assets:** Upgrade Inter font to version 4.1 (bedrock #15999)
```
maureenlholland, line 5: needs to be rebased and moved above release 19.3.0

maureenlholland (approved):
r+wc

after latest procotol release, changelog needs a minor update

CHANGELOG.md
janbrasna: Q: Do the WOFF removals (now only WOFF2 left) need any migration notes, regarding the change in browser support? https://caniuse.com/woff2 vs. https://caniuse.com/woff

janbrasna: (esp. since [#1030](https://github.com/mozilla/protocol/pull/1030/files) still seems to include WOFF 1.0 for future releases.)

maureenlholland: Good eye. I'm inclined to add migrate note and use woff2 only for new fonts, but have asked internally as well.

maureenlholland: We're going to go ahead with woff2 only. It's well supported and the fallback experience is acceptable for degraded support. I'll update the Mozilla typeface PR to remove woff.

@wen-2018 one last changelog request: we should call out that we are dropping support for `woff`

PR #1034, opened 2025-03-14
Remove unused /bin CI scripts

janbrasna:
## Description

These come from times of Jenkins and GitLab CI, with S3 as the deployment — none of which is currently in use. Site is currently built at Netlify via hooks/oauth app plugged directly into VCS/GH, both for _prod_ and _stage_ (and demo branches as well), following its config toml and manifest scripts exclusively.

### Testing

Not used most likely even elsewhere less visible, as the `node:16` image would be complaining about compatibility otherwise, as deps are already bumped beyond that.

(While I don't have `#www-notify` access, my guess is the deployment status from MZP is not being posted there anyways, at least not via this script, as it expects `slack-cli` on the runner which probably was configured on moz.works blue CI Jenkins instance, but here it's most likely not set up, incl. any secrets, right? Any current integration would use `slackapi/slack-github-action` instead, with `SLACK_BOT_TOKEN` configured in the env secrets for more native GHA approach…)


stevejalim (approved):
Yeah I can't see how slack is pinged from here, but it doesn't look like it's from the shell script. R+ and 🤞

stevejalim:
@maureenlholland Feel free to merge whenever it suits you

PR #1042, opened 2025-03-18
Add pointer to ‹details› element ‹summary›

### LINKED ISSUE #948: Apply hover cursor from Details component to Details element too
## Description

Details element https://protocol.mozilla.org/components/detail/details, when compared to Details component https://protocol.mozilla.org/components/detail/details-component--default, doesn't offer the same `cursor: pointer` that improves its usability/affordance.

The current state perfectly matches how a `<details>` element behaves in default browser styles, which is somewhat sub-par if not styled better. (The plaintext selection caret may obscure the fact such element is to be interacted with.)

### Steps to reproduce

Using only basic `<details>` instead of the more complex `mzp-c-details`, ref: https://github.com/mozilla/bedrock/issues/14781

### Expected result

Interactive mouse pointer signaling the UI element has a function, to match it with the Details Protocol component.

### Actual result

Only default pointer affordance (text selection, caret) styling on hover.

(UA defaults: https://stackoverflow.com/a/77766654; styling tips: https://css-tricks.com/two-issues-styling-the-details-element-and-how-to-solve-them/#aa-changing-the-summary-cursor-value incl. `@supports not (-ms-ime-align: auto)` for not applying it when not interactive.)

### Environment

macOS Catalina
FXDX 128b8, Chromium 127r1313161

janbrasna:
## Description

Details _element_ gets styled very close to Details _component_ with many of the enhancements, just the hover affordance is missing for this plan HTML5 use case. This sets the cursor to pointer, as the full component does.

### Issue

Resolves #948

### Testing

https://patch-2--mzp-dev.netlify.app/components/detail/details

1. I have not fenced this behind any `@supports not (-ms-ime-align: auto)` of sorts for things like IE that don't provide the functionality as that adds extra cruft (and [false positives](https://github.com/mozilla/protocol/issues/948#issuecomment-2218321222)) for not much benefit, when MzpDetails actually provides a polyfill for such browsers.
2. I have opted for a plain extra declaration instead of the `@summary` mixin (which gets included within half a dozen of things) not to apply such styling to places that render it own ways.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Features
 
+* **css:** Apply hover cursor from Details component to Details element (#948)
+
 ## Bug Fixes
 
 * **assets:** Update @mozilla-protocol/assets to 5.4.0
--- a/assets/sass/protocol/base/elements/_details.scss
+++ b/assets/sass/protocol/base/elements/_details.scss
@@ -30,6 +30,10 @@ details .is-summary button[aria-expanded='true']::before,
     @include summary-open;
 }
 
+summary  {
+    cursor: pointer;
+}
+
 summary::-webkit-details-marker {
     display: none;
 }


stephaniehobson (approved):
R+ ☝️

PR #1043, opened 2025-03-20
Remove all mixins supporting vendor prefixes (Fix #957)

### LINKED ISSUE #957: Audit use of vendor prefix mixins
There are a handful of utility mixins sometimes serving "interesting" prefixed rules:

https://github.com/mozilla/protocol/blob/12419244bf58ae0237edf062983c2852b56741e3/assets/sass/protocol/includes/mixins/_utils.scss#L101-L114

(e.g. `align-items` is unprefixed since Safari 9, and Safari 8 had it prefixed, but the fallback to `*-box-align` i.e. another experimental property sounds rather complex for today's use; `-webkit-background-size` had a different logic applied to param count and didn't understand a handful of values etc.)

While definitely intriguing piece of code that must have been really helpful in the past decade, some bits are probably no longer relevant or necessary, and if used, only inflating the compiled output with no positive impact. Maybe it's time to audit if/where these are being used, and plan a decommission path to have them removed one day.

If a major/breaking release is coming up, that might help justifying some of the mixin removal, if they no longer serve the purpose and would just need to be kept around passing through the stable property for backwards compatibility of the API.

stephaniehobson:
## Description

This version removes mixins which added vendor-prefixes. Browser support for these is now excellent.

- One migration path is to edit your code to use the unprefixed versions.

### Issue

Fix #957

### Testing

View the affected components to make sure they are still displaying properly.


janbrasna (commented):
💟 Love it!

Unfortunately I can't seem to identify the `mzp-a-fade-*` build issue just from glancing it, pretty sure it's all in the correct order… 🤷 

Some copy/formatting nits in the meantime:

stephaniehobson:
I'm pretty sure the `mzp-a-fade-*` build issue has turned up because I'm actually declaring an animation in the code now - instead of what we were doing before which was use an include that was not being evaluated for validity. So, a nice side affect of the change is better linting! The downside is I think we have to disable the rule since Stylelint doesn't evaluate Sass includes/use rules.

maureenlholland:
Do Not Merge until Icon release is published

stephaniehobson:
I ran some visual regression tests against this branch locally and no differences were found in Chrome.

PR #1044, opened 2025-03-28
Draft PR for discussion: transition to CSS varaibles only.

stephaniehobson:
Draft PR for discussion.

### DIFF
--- a/assets/sass/protocol/base/utilities/_titles.scss
+++ b/assets/sass/protocol/base/utilities/_titles.scss
@@ -10,73 +10,41 @@
 .mzp-u-title-2xl {
     @include text-title-2xl;
     font-family: $title-font-family;
-
-    @supports (--css: variables) {
-        font-family: var(--title-font-family);
-    }
 }
 
 .mzp-u-title-xl {
     @include text-title-xl;
     font-family: $title-font-family;
-
-    @supports (--css: variables) {
-        font-family: var(--title-font-family);
-    }
 }
 
 .mzp-u-title-lg {
     @include text-title-lg;
     font-family: $title-font-family;
-
-    @supports (--css: variables) {
-        font-family: var(--title-font-family);
-    }
 }
 
 .mzp-u-title-md {
     @include text-title-md;
     font-family: $title-font-family;
-
-    @supports (--css: variables) {
-        font-family: var(--title-font-family);
-    }
 }
 
 .mzp-u-title-sm {
     @include text-title-sm;
     font-family: $title-font-family;
-
-    @supports (--css: variables) {
-        font-family: var(--title-font-family);
-    }
 }
 
 .mzp-u-title-xs {
     @include text-title-xs;
     font-family: $title-font-family;
-
-    @supports (--css: variables) {
-        font-family: var(--title-font-family);
-    }
 }
 
 .mzp-u-title-2xs {
     @include text-title-2xs;
     font-family: $title-font-family;
-
-    @supports (--css: variables) {
-        font-family: var(--title-font-family);
-    }
 }
 
 .mzp-u-title-3xs {
     @include text-title-3xs;
     font-family: $title-font-family;
-
-    @supports (--css: variables) {
-        font-family: var(--title-font-family);
-    }
 }
 
 // Utility class for centered text.
--- a/assets/sass/protocol/includes/mixins/_typography.scss
+++ b/assets/sass/protocol/includes/mixins/_typography.scss
@@ -8,6 +8,7 @@
 @use '../themes-sass' as themes;
 @use '../functions';
 
+
 // * -------------------------------------------------------------------------- */
 // Font families
 @mixin font-base {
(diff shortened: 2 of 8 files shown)


assets/sass/protocol/base/utilities/_titles.scss
```diff
@@ -10,73 +10,41 @@
 .mzp-u-title-2xl {
     @include text-title-2xl;
     font-family: $title-font-family;
```
stephaniehobson, line 12: Maybe font-family should be part of the `@include text-title-n` declaration?

assets/sass/protocol/includes/themes/_firefox.scss
```diff
@@ -35,51 +35,101 @@
...
-    --title-lg-line-height: 1.166;
-    --title-md-size: 2.5rem;
-    --title-md-line-height: 1.1;
-    --title-sm-size: 2rem;
-    --title-sm-line-height: 1.125;
-    --title-xs-size: 1.5rem;
-    --title-xs-line-height: 1.166;
-    --title-2xs-size: 1.25rem;
-    --title-2xs-line-height: 1.2;
+    --title-2xl-size: 3.5rem;
```
stephaniehobson, line 38: I've been toying with the idea of variable names that make it clear that they morph with media queries. Alternative name ideas:

- `--scale-title-2xl-size: 3.5rem;`(colours could use `theme`, nice because it gives a clue of the variable value)
- `--let-title-2xl-size: 3.5rem;` (javascript reference: const vs let, mostly nice because its short, we should not have any const CSS vars, they should be Sass vars)
- `--find-title-2xl-size: 3.5rem;` (nice because all dynamic variables can use the same prefix)

assets/sass/protocol/includes/mixins/_typography.scss
```diff
@@ -55,223 +55,68 @@
...
-    @supports (--css: variables) {
-        font-size: var(--title-xl-size);
-        line-height: var(--title-xl-line-height);
-
-        @media #{tokens.$mq-md} {
-            font-size: var(--title-2xl-size);
-            line-height: var(--title-2xl-line-height);
-        }
-    }
+    font-size: var(--title-2xl-size);
```
stephaniehobson, line 58: Would putting the CSS variable into a Sass variable help with readability and authoring speed? Or would it add to confusion?

It would be SO NICE to just type `--title-2xl-size`.

maureenlholland:
For color naming, I think it's great to have variables that can react to both theme (fx/moz) and color-scheme (light/dark).

We'll need to update m24 naming to match Protocol's usage-based names (like `--text-title-color`). Is there anything we'd like to change about the current color usage-based names (other than potential prefix)?
inverted-colors

light-dark

maureenlholland:
**(Might be out of scope for discussion)**

Could we replace the hover colors with a function? (Sass darken or CSS color-mix or relative colors) Two fewer tokens to maintain, easier to scale a consistent hover style if we add other link colors later
utility-colors

assets/sass/protocol/includes/mixins/_typography.scss
```diff
@@ -38,7 +39,6 @@
 
     @if math.unit($size) == 'px' {
         font-size: $size;
-        font-size: functions.remify($size);
```
janbrasna, line 41: I'm wondering whether the idea wasn't to actually remove the other line — to drop the px fallback output — as now all it does is just a pass-through for px values (and an error for all other cases), and while `font-size()` mixin is no longer used here below favoring just vars, it's still being used e.g. by the button component feeding it px values to presumably remify.

janbrasna (commented):
The failed check is just a linter being loud about whitespace, nothing more.

Couple of naive questions follow, esp. if the change to the px-to-rem mixin is intentional:

assets/sass/protocol/includes/themes/_firefox.scss
maureenlholland: Minor preferences:
- scale/theme prefixes on CSS custom properties look good to me, I like the extra indication of change based on context
- `var()` syntax doesn't bother me but happy to save a few characters with alias to Sass variables

assets/sass/protocol/includes/mixins/_typography.scss
stephaniehobson: Turns out this is a reserved namespace in Sass.

assets/sass/protocol/includes/themes/_mozilla.scss
```diff
@@ -85,7 +85,7 @@
         --body-md-size: 1rem;
         --body-sm-size: 0.875rem;
         --body-xs-size: 0.75rem;
-        --body-line-height: 1.2;
+        // --body-line-height: 1.2; doesn't change, no need to update
```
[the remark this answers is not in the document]
stephaniehobson: I've omitted all the unnecessary re-definitions in both Mozilla and Firefox theme files instead of commenting them out. The code is cleaner, hopefully it doesn't confuse anyone.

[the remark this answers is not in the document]
stephaniehobson: Nice catch, thanks ;)

assets/sass/protocol/includes/mixins/_typography.scss
stephaniehobson: Yup 😳

Issue #758, opened 2021-12-16
Modal styles assume light color themed links


alexgibson:
## Description

The current modal styles seem to assume that all links contained within the modeal will be light color themed. This can cause issues on links that appear on dark backgrounds, or use colored buttons.

Ref: https://github.com/mozilla/protocol/blob/main/src/assets/sass/protocol/components/_modal.scss#L105

### Steps to reproduce

See https://github.com/mozilla/bedrock/issues/9866

### Expected result

We probably shouldn't make an assumption about the content within the modal?

### Actual result

It is assumed that links will all appear on a light background.

### Environment

Firefox 95
macOS

janbrasna:
Currently the Modal component in Protocol is designed to be a black panel. It then `@includes` light-links which sets: https://github.com/mozilla/protocol/blob/2da9964461a66eabef3f5b0de0172a4575e91450/assets/sass/protocol/includes/_themes-sass.scss#L43-L48

so any consumer changing the modal to white also has to override all the link pseudos to get a readable result.

Probably the best solution would be to allow both light and dark themes for the whole component, which would only set the light-links for the dark appearance, and keeping the defaults otherwise, inheriting their styles.

janbrasna:
Raised before as part of #855 in #357

Issue #655, opened 2021-01-22
[a11y] Button color contrast issue for focus / visited link styles


alexgibson:
## Description

For links that are styled like buttons, the focus text color is hard to read if the link destination has already been visited.

### Steps to reproduce

1.) Visit https://www.mozilla.org/en-US/about/
2.) Click on the "Read our Mission" cta button.
3.) Once on the mission page, press the browser back button.
4.) Now using the keyboard tab to focus on the cta button again.

### Expected result

The button text should be readable.

### Actual result

The button text has contrast issues:

<img width="841" alt="image" src="https://user-images.githubusercontent.com/400117/105499722-ccd2bc00-5cb9-11eb-8276-e9baa021bf89.png">

### Environment

Firefox Nightly

PR #1045, opened 2025-03-29
Fix contrast of visited/focus Button links

### LINKED ISSUE #842: Button link focus styling doesn't override inherited visited link styling
## Description

When a link is styled as a button it can take on an additional `:visited` state that `button` elements don't get. When a button link has been visited and is then focused, the regular visited link colors still apply and the button becomes unreadable.

image

Button focus styles should be consistent whether it's a link or a `button` so we may need some additional rules to handle `:visited:focus`.

janbrasna:
## Description

Adds a little more specificity necessary to not inherit own anchor styles for better resilience.

### Issue

Fixes #842

### Testing

Current styles with added "visited" state by converting the buttons to anchors linking the same preview page recursively, to demonstrate the underlying issue more visibly:
https://control--mzp-dev.netlify.app/components/detail/button--using-link

PR preview:
https://cta--mzp-dev.netlify.app/components/detail/button--using-link


components/button/button.config.yml
```diff
@@ -2,49 +2,61 @@ context:
...
+    notes: CTA anchor styled as Button.
+    context:
+      link: './button--using-link'
   - name: Primary Dark
     preview: '@preview-dark'
     notes: Add the class `mzp-t-dark` for use on dark backgrounds (the button itself is light).
     context:
       class: 'mzp-t-dark'
+      link: './button--primary-dark'
   - name: Secondary
```
janbrasna, line 15: I have changed the examples from buttons to links to better show any link-related woes right in the Fractal preview. Originally it was only for debug purposes, but I'm thinking about keeping it that way to help surface any anchor pseudo issues.

(That meant adding the anchor as another variant to keep the default button — as form-related styling still has to be demonstrated and some states as readonly/disabled example are derived from it. All the other examples were just switched to links without any further impact, didn't want to duplicate these.)

components/button/button.config.yml
```diff
@@ -2,49 +2,61 @@ context:
   label: Primary Button
 default: Primary
 variants:
+  - name: Using link
```
maureenlholland, line 5: 👍

components/button/button.config.yml
maureenlholland: makes sense to me, glad we have both elements in preview

maureenlholland (approved):
Good to merge once the changelog is updated r+wc (edit: this can wait until the notification update is merged!)

I also left a comment on the bigger change I'm interested in exploring for links within the design system. Not something to address in this PR, just thinking out loud

assets/sass/protocol/components/_button.scss
```diff
@@ -138,9 +134,9 @@ a.mzp-c-button {
 // * -------------------------------------------------------------------------- */
 // Primary
 
-/* stylelint-disable-next-line no-duplicate-selectors */
 .mzp-c-button,
-a.mzp-c-button {
+a.mzp-c-button:link,
```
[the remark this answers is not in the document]
stephaniehobson: Oh yeah, the specificity of link styles is a constant problem 😭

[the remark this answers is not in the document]
janbrasna: Changelog updated, should not be conflicting with the pending v21 changes as well, so should merge cleanly in any order.

(I'll bring the specificity conversation back to the tracking issue to follow up, but just for the context of this PR: There was this issue back in the day #655 that got resolved, but later disappeared in [#864](https://github.com/mozilla/protocol/pull/864/files) — while investigating the referenced #842 issues, I actually only ran into tightly scoped regressions, or implementation bugs — like #758 where light-links or white-links are part of the component, but the consumer never changed that etc. — so the only real issue, before having the M24 base in bedrock, that does indeed cause a lot of headache redefining both MZP themes with higher specificity, was the actual bug addressed here in the PR, i.e. the "primary" and "primary-dark" button themes not having the necessary color explicitly. Not that a more modern approach to allowing greater flexibility is not needed today, esp. with moving to CSS vars fully, just that it might be worth mentioning all the individual occurrences mentioned never really shown a systemic issue, but always ended up being very concrete bugs instead. _NB: From my investigations, yea, color math, where and finding a way to replace wholesale link color mixins all sound like a great improvement!_)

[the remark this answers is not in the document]
janbrasna: (Changelog updated from release to head again, conflicts resolved.)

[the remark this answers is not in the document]
stephaniehobson: I think maybe you didn't push the rebase? This note is from today but the last update is 5 days ago and I still see a conflict 🙃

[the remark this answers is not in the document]
janbrasna: @stephaniehobson This is weird. Was the same on previous PR with Alex — no idea why it's all green from my side but the last push doesn't appear for a maintainer… I do not rebase unless strictly needed (not to rewrite anything that's been already reviewed, and also to allow reviewing just the conflict resolution merge by itself if needed); for mere changelog conflicts I keep using the PR UI to edit and mark as resolved, i.e. adding in merge commits to update. At the point of the comment, a 5 days old change would be outdated, there was a newer merge 3 days old (I even added extra time to update any caches or stuck state) and there are CI run logs matching that; it's just weird. I pushed another update right now, so, YMMV, but, again updated on last changelog update there;)

stephaniehobson:
Finally green and merged 🙂

assets/sass/protocol/components/_button.scss
```diff
@@ -69,10 +69,6 @@
     }
 }
 
-@mixin product-states-dark {
-    @include product-states;
-}
-
 // * -------------------------------------------------------------------------- */
 // Button shape and size
```
janbrasna, line 73: Thinking of this now… I haven't really brought up whether it's okay to just drop this mixin without any breaking notice or migration hints — from what I understand it was only ever used internally to style the individual classes of themes of buttons, and never really exposed to consumers per se… but who knows in times of everything being available in the global namespace, right? 😇

So, if anyone could help me judge this… does it look like an internal shim from way back, and is safe to drop silently, or should be noted for consumers to be on the safe side?

stephaniehobson: A quick GitHub search suggests it's not used anywhere. I'd still note it in the CHANGELOG to be on the safe side though.

janbrasna: OK, I actually have a draft backfilling some missing entries overall, so I'll expedite that work and add this as well.

_(Or, this can just be reverted and kept for posterity; now I see that when it was added in [#562/files](https://github.com/mozilla/protocol/pull/562/files#diff-af9a2d7922d600b0eacaad183de85ee7795c7d4e9022dc952effecc85bdd4ee9) it was merely for consistency anyways?…)_

PR #1054, opened 2025-04-16
[WIP] Reduce base link specificity 

maureenlholland:
## Description

This is more a thought experiment than anything at this stage. The idea is to reduce base element link specificity so it is easily overridden by a single class

Base element specificity: 0 0 1
not visited
Screenshot 2025-04-16 at 12 33 56 PM

visited
Screenshot 2025-04-16 at 12 32 03 PM

Class specificity: 0 1 0

NOTE: the danger of custom properties is that they inherit all the way down the tree, so could be going further than desired if "donut scope" was intended

### Issue

Add a link to a related GitHub issue if applicable.

### Testing

### DIFF
--- a/assets/sass/protocol/base/elements/_links.scss
+++ b/assets/sass/protocol/base/elements/_links.scss
@@ -4,12 +4,6 @@
 
 @use '../../includes/lib' as *;
 
-// check support and fallback of :where
-a:where(:visited) {
-    --link-color: var(--link-color-visited);
-    --link-color-hover: var(--link-color-visited-hover);
-}
-
 a {
     color: var(--link-color);
     text-decoration: underline;
@@ -21,6 +15,14 @@ a {
 
     &:where(:active) {
         background-color: rgba(0, 0, 0, 0.05);
+    }
+}
+
+a:where(:visited) {
+    color: var(--link-color-visited);
+
+    &:where(:hover, :focus, :active) {
+        color: var(--link-color-visited-hover);
     }
 }
 
--- a/assets/sass/protocol/base/elements/_links.scss
+++ b/assets/sass/protocol/base/elements/_links.scss
@@ -4,61 +4,42 @@
 
 @use '../../includes/lib' as *;
 
-:link {
-    color: $link-color;
+a {
+    color: var(--link-color);
     text-decoration: underline;
 
     &:hover,
     &:focus,
     &:active {
-        color: $link-color-hover;
+        color: var(--link-color-hover);
         text-decoration: none;
     }
 
     &:active {
         background-color: rgba(0, 0, 0, 0.05);
     }
-
-    @supports (--css: variables) {
-        color: var(--link-color);
-
-        &:hover,
-        &:focus,
-        &:active {
-            color: var(--link-color-hover);
-        }
-    }
 }
 
 :visited {
-    color: $link-color-visited;
-
-    &:hover,
-    &:focus,
-    &:active {
-        color: $link-color-visited-hover;
-    }
-
-    @supports (--css: variables) {
-        color: var(--link-color-visited);
-
-        &:hover,
-        &:focus,
-        &:active {
-            color: var(--link-color-visited-hover);
-        }
-    }
+    --link-color: var(--link-color-visited);
+    --link-color-hover: var(--link-color-visited-hover);
 }
 
 .mzp-t-dark {
-    @include light-links;
+    --link-color: var(--link-color-inverse);
+    --link-color-hover: var(--link-color-hover-inverse);
+    --link-color-visited: var(--link-color-visited-inverse);
+    --link-color-visited-hover:var(--link-color-visited-hover-inverse);
+}
+
+.mzp-t-white-links {
+    --link-color: $color-white;
+    --link-color-hover: $color-white;
+    --link-color-visited: $color-white;
+    --link-color-visited-hover: $color-white;
 }
 
 .mzp-c-cta-link {
-    font-family: $button-font-family;
+    font-family: var(--button-font-family);
     font-weight: bold;
-
-    @supports (--css: variables) {
-        font-family: var(--button-font-family);
-    }
 }
--- a/assets/sass/protocol/base/elements/_links.scss
+++ b/assets/sass/protocol/base/elements/_links.scss
@@ -4,6 +4,12 @@
 
 @use '../../includes/lib' as *;
 
+// check support and fallback of :where
+a:where(:visited) {
+    --link-color: var(--link-color-visited);
+    --link-color-hover: var(--link-color-visited-hover);
+}
+
 a {
     color: var(--link-color);
     text-decoration: underline;
@@ -15,12 +21,6 @@ a {
 
     &:where(:active) {
         background-color: rgba(0, 0, 0, 0.05);
-    }
-
-    // check support and fallback of :where
-    &:where(:visited) {
-        --link-color: var(--link-color-visited);
-        --link-color-hover: var(--link-color-visited-hover);
     }
 }
 
@@ -36,10 +36,10 @@ a {
 // any descendant link will inherit white theming (replace white-links mixin)
 // link-specific theme class
 .mzp-t-white-links {
-    --link-color: $color-white;
-    --link-color-hover: $color-white;
-    --link-color-visited: $color-white;
-    --link-color-visited-hover: $color-white;
+    --link-color: #{$color-white};
+    --link-color-hover: #{$color-white};
+    --link-color-visited: #{$color-white};
+    --link-color-visited-hover: #{$color-white};
 }
 
 // this might belong in a different CTA file?
--- a/assets/sass/protocol/base/elements/_links.scss
+++ b/assets/sass/protocol/base/elements/_links.scss
@@ -8,21 +8,19 @@ a {
     color: var(--link-color);
     text-decoration: underline;
 
-    &:hover,
-    &:focus,
-    &:active {
+    &:where(:hover, :focus, :active) {
         color: var(--link-color-hover);
         text-decoration: none;
     }
 
-    &:active {
+    &:where(:active) {
         background-color: rgba(0, 0, 0, 0.05);
     }
-}
 
-:visited {
-    --link-color: var(--link-color-visited);
-    --link-color-hover: var(--link-color-visited-hover);
+    &:where(:visited) {
+        --link-color: var(--link-color-visited);
+        --link-color-hover: var(--link-color-visited-hover);
+    }
 }
 
 .mzp-t-dark {
--- a/assets/sass/protocol/base/elements/_links.scss
+++ b/assets/sass/protocol/base/elements/_links.scss
@@ -17,12 +17,15 @@ a {
         background-color: rgba(0, 0, 0, 0.05);
     }
 
+    // check support and fallback of :where
     &:where(:visited) {
         --link-color: var(--link-color-visited);
         --link-color-hover: var(--link-color-visited-hover);
     }
 }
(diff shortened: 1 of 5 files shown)


maureenlholland:
Hoping this direction is in line with [transition to CSS variables](https://github.com/mozilla/protocol/pull/1044)

assets/sass/protocol/base/elements/_links.scss
```diff
@@ -4,61 +4,50 @@
 
 @use '../../includes/lib' as *;
 
-:link {
-    color: $link-color;
+a {
+    color: var(--link-color);
```
maureenlholland, line 8: This was interesting. I thought I could redeclare the `--link-color` var within the `:visited` selector scope, but it didn't seem to take effect as computed style. Dev tools seems to show both the correct variable and the wrong computed style 🤷‍♀️ 

```
:root {
  --link-color: default-color;
  --link-color-visited: visited-color;
}

a {
  color: var(--link-color);
}

a:where(:visited) {
  --link-color: var(--link-color-visited);
}

Comment (maureenlholland) [reply assets/sass/protocol/base/elements/_links.scss:8]: pseudo-class variable correctly identified
computed value still default color
Screenshot 2025-04-16 at 12 13 36 PM

root inheritance correctly overruled
Screenshot 2025-04-16 at 12 14 14 PM

File: assets/sass/protocol/base/elements/_links.scss [hunk 2]
```diff
@@ -4,61 +4,50 @@
...
-            color: var(--link-color-visited-hover);
-        }
+    &:where(:hover, :focus, :active) {
+        color: var(--link-color-visited-hover);
     }
 }
 
+// any descendant link will inherit dark theming (replace light-links mixin)
+// this might belong in root file? site-wide dark theme class
 .mzp-t-dark {
```
maureenlholland, line 31: inherited values
Screenshot 2025-04-16 at 12 35 52 PM

assets/sass/protocol/base/elements/_links.scss
```diff
@@ -4,61 +4,50 @@
...
-    @include light-links;
+    --link-color: var(--link-color-inverse);
+    --link-color-hover: var(--link-color-hover-inverse);
+    --link-color-visited: var(--link-color-visited-inverse);
+    --link-color-visited-hover:var(--link-color-visited-hover-inverse);
+}
+
+// any descendant link will inherit white theming (replace white-links mixin)
+// link-specific theme class
+.mzp-t-white-links {
```
maureenlholland, line 40: Inherited values
Screenshot 2025-04-16 at 12 37 39 PM

assets/sass/protocol/base/elements/_links.scss
```diff
@@ -4,61 +4,50 @@
...
-        color: var(--link-color-visited);
+a:where(:visited) {
+    color: var(--link-color-visited);
 
-        &:hover,
-        &:focus,
-        &:active {
-            color: var(--link-color-visited-hover);
-        }
+    &:where(:hover, :focus, :active) {
```
maureenlholland, line 24: wiping out specificity of stateful pseudo-classes with `:where()` 
Screenshot 2025-04-16 at 12 37 28 PM

assets/sass/protocol/base/elements/_links.scss
```diff
@@ -4,61 +4,50 @@
...
 }
 
+// any descendant link will inherit dark theming (replace light-links mixin)
+// this might belong in root file? site-wide dark theme class
 .mzp-t-dark {
-    @include light-links;
+    --link-color: var(--link-color-inverse);
+    --link-color-hover: var(--link-color-hover-inverse);
+    --link-color-visited: var(--link-color-visited-inverse);
+    --link-color-visited-hover:var(--link-color-visited-hover-inverse);
```
janbrasna, line 35:
```suggestion
    --link-color-visited-hover: var(--link-color-visited-hover-inverse);
```

janbrasna (commented):
This makes sense, looks elegant, and is easy to follow/understand. (Redefining the vars for states as shown above resulting in the rendering woes looks cool at first, but this take looks more reliable.)

The tricky part might be following the themes to the actual components, to make sure it either still works as intended, or what tweaks could be useful there… as it's usually the components being consumed and redefined downstream causing the specificity issues, e.g. the Modal: (#758) https://github.com/mozilla/protocol/blob/d34e087c4a0306728564506abe9810cb90cf17b2/assets/sass/protocol/components/_modal.scss#L104-L106

assets/sass/protocol/components/_button.scss
```diff
@@ -76,8 +76,7 @@
 // * -------------------------------------------------------------------------- */
 // Button shape and size
 
-.mzp-c-button, /* stylelint-disable-line no-duplicate-selectors */
-a.mzp-c-button {
+.mzp-c-button /* stylelint-disable-line no-duplicate-selectors */ {
```
janbrasna, line 79:
```suggestion
.mzp-c-button {
```

assets/sass/protocol/components/_button.scss
```diff
@@ -139,8 +138,7 @@ a.mzp-c-button {
 // Primary
 
 /* stylelint-disable-next-line no-duplicate-selectors */
```
janbrasna, line 140:
```suggestion
```

janbrasna (commented):
These should also no longer complain:

assets/sass/protocol/components/_button.scss
maureenlholland: We do still have a duplicate selector here (there's no `mzp-t-primary` theme class to distinguish this selector from the other default button styles), but this is the only place we need the disable comment now 

duplicate-selector

assets/sass/protocol/base/elements/_links.scss
stephaniehobson: I love the move the variables here! We need to be careful we also move to variables for the background colours _at the same time_ so we don't get in a circumstance where the background is defined as dark using a hex colour but the component depends on variables for the links.

Perhaps the prudent thing to do is to bundle the CSS variable changes into one big point release version that we test on un-supporting browsers.

assets/sass/protocol/base/elements/_links.scss
stephaniehobson: I'm worried about using `:where` since the support for it is less wide spread than support for dynamic properties. Can you help me understand what would happen if a visitor had variable support but not where support? Links are visible they just don't get hover/focus? Or something more dire?

assets/sass/protocol/base/elements/_links.scss
stephaniehobson: Another thing we could do to make styling links easier would be to drop the separate styles for visited links on hover/focus/active and [rely on source order for the override](https://css-tricks.com/css-basics-styling-links-like-boss/#aa-style-each-link-state).

1.     Link
2.     Visited
3.     Hover
4.     Active
5.     Focus

stephaniehobson: I've always found the :link:visited:hover edge case to be confusing.

assets/sass/protocol/components/_button.scss
```diff
@@ -160,32 +158,27 @@ a.mzp-c-button {
...
 // Secondary
 
-/* stylelint-disable-next-line no-duplicate-selectors */
-.mzp-c-button,
-a.mzp-c-button {
-    &.mzp-t-secondary {
-        background-color: transparent;
-        border-color: $color-black;
-        color: $color-black;
+.mzp-c-button.mzp-t-secondary {
```
stephaniehobson, line 161: This is so much nicer, I want this to work 😭

assets/sass/protocol/includes/themes/_firefox.scss
```diff
@@ -25,7 +25,7 @@
     --link-color-hover: #{tokens.$color-link-hover};
     --link-color-inverse: #{tokens.$color-blue-10};
     --link-color-visited-hover-inverse: #{tokens.$color-violet-05};
-    --link-color-visited-hover: #{tokens.$color-link-hover};
+    --link-color-visited-hover: #{tokens.$color-link-visited-hover};
```
stephaniehobson, line 28: Another place that would be simplified by having the same hover state for both visited and unvisited links.

assets/sass/protocol/base/elements/_links.scss
stephaniehobson: I looked into [CSS layers as a possible solution ](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer) (I am so excited for these!) but the support for them is about the same as support for :where()

assets/sass/protocol/base/elements/_links.scss
maureenlholland: Will add a commit with source order for comparison

assets/sass/protocol/base/elements/_links.scss
```diff
@@ -4,61 +4,51 @@
...
-    &:focus,
-    &:active {
-        color: $link-color-hover;
+    &:visited {
+        color: var(--link-color-visited);
+    }
+
+    &:hover, &:active, &:focus {
+        /* stylelint-disable-next-line color-function-notation */
+        color: var(--link-color-with-state, hsl(from var(--link-color) h s calc(l - 10)));
```
maureenlholland, line 24: Demo of functional hover color approach:

1. We have a default that automatically darkens the lightness of the color by 10% 
2. We can override this default with another function (i.e. dark theme automatically lightens color) or another set value (i.e. white theme uses white)

maureenlholland: We could also do this with sass `color.scale` for better support but we have to use a sass variable as the input argument (CSS custom properties are not available as values at build time)
need-sass-variable

assets/sass/protocol/base/elements/_links.scss
maureenlholland: The only potential conflict here is that a stateful pseudo-class + element selector will beat a component class selector (if it doesn't also explicitly override the state). This is most likely to happen with `button` component and link-specific `visited` state.

We could unnest the link states so they are only 0,1,0 specificity and rely on source order to override (basic element styles are overruled by component styles)

Or we could also include the `:visited` state selector on any elements that could also be links under the hood

assets/sass/protocol/base/elements/_links.scss
maureenlholland: This would allow us to drop support for `hover` tokens and (hopefully) simplify maintenance

assets/sass/protocol/base/elements/_links.scss
stephaniehobson: So on a background set to black with a CSS variable they will get blue/purple links?

maureenlholland: Will check tomorrow. I think they will get the `--link-color`or `--link-color-visited` value. If we use `mzp-t-dark` for dark background, the links will still have good contrast (those variables will respond to the dark theme), but their color will not change on hover/focus.

maureenlholland: Ok, so older browser links would always get the `--link-color` regardless of state or `:visited` status (all those rules are ignored). The correct link color will apply according to light/dark theme.

states
visited

maureenlholland:
After internal discussion, we want to [drop support for `:hover:visited` style](https://github.com/mozilla/protocol/issues/1071) and rely on order of appearance to ensure `:hover` style always applied on hover state (whether link is visited or not)

Do Not Merge until Icon release is published

stephaniehobson:
Closing in favour of #1077

PR #1077, opened 2025-07-23
Remove :hover:visited styles (Fix #1071)

### LINKED ISSUE #1071: Remove :hover:visited styles
We want to keep `:visited` and `:hover`, but we don't want to maintain the combination of the two.

Hover will take precedence over visited (there will be no specific Visited Hover color)

stephaniehobson:
## Description

### Issue

Add a link to a related GitHub issue if applicable.

### Testing


assets/sass/protocol/components/_button.scss
```diff
@@ -135,8 +135,7 @@ a.mzp-c-button {
 // Primary
 
 .mzp-c-button,
-a.mzp-c-button:link,
-a.mzp-c-button:visited {
+a.mzp-c-button {
```
stephaniehobson, line 138: The increased specificity was added in https://github.com/mozilla/protocol/pull/1045 to address the problem where links were styled like buttons and then got focus.

This has been solved by removing the text colour change for :focus styles from the default :link styles.

assets/sass/protocol/base/elements/_links.scss
```diff
@@ -8,8 +8,11 @@
     color: $link-color;
     text-decoration: underline;
 
+    &:visited {
```
stephaniehobson, line 11: `:visited` moves earlier in the source so is over-ridden by the hover, focus, and active styles that follow. We rely on the cascade to style a link when it's hovered instead of explicitly declaring :visited:hover styles.

assets/sass/protocol/components/_footer.scss
```diff
@@ -25,12 +25,12 @@
...
         font-weight: normal;
         text-decoration: underline;
+    }
 
-        &:hover,
-        &:focus,
-        &:active {
-            text-decoration: none;
-        }
+    a:hover,
```
stephaniehobson, line 30: Another place to decrease specificity and rely on the cascade.

components/button/button.html
```diff
@@ -9,6 +9,33 @@
 # iconStart     - markup
 # iconEnd       - markup
 #}
+
```
stephaniehobson, line 12: Testing code - should be removed before the PR is opened for review.

assets/sass/protocol/base/elements/_links.scss
```diff
@@ -8,8 +8,11 @@
...
+        color: $link-color-visited;
+    }
+
     &:hover,
-    &:focus,
```
stephaniehobson, line 12: Declaring the colour change on focus was the root problem in https://github.com/mozilla/protocol/issues/842

stephaniehobson:
Note to self for next week: This is still a draft because there's some testing HTML to remove and because you want to do more testing on bedrock.

Issue #982, opened 2024-09-17
Go all-in on CSS custom properties for themes, remove Sass variables


craigcook:
Much of the preliminary work we've done in bedrock for the new Mozilla branding has hinged on custom properties and eschewed Sass variables. And frankly it simplified SO MUCH. It makes theme-switching and color modes much easier to implement without worrying about Sass compile errors or importing things in the right order.

We [semi-recently converted Protocol to use custom properties for theming](https://github.com/mozilla/protocol/commit/e244b27165e6ef16bc4a7e47620091747940da05) but preserved the Sass variables as fallbacks for older browsers. But looking at [support levels and browser usage stats,](https://caniuse.com/css-variables) I've been thinking it's time to embrace the future of CSS and remove those legacy Sass variables (at least for all the theme-level properties; some things can/should remain as Sass variables and mixins).

Issue #668, opened 2021-03-16
Rename "title" to "heading"


craigcook:
We use the term "title" extensively, in components, mixins, variables, and tokens. "Heading" is arguably a better term because it matches the HTML element the term almost always applies to. It also helps to distinguish it from other uses of "title" in HTML (the `title` attribute and the `title` element in the document head).

Changing this term will be pretty sweeping and is certainly a breaking change. We can alias some instances to try to reduce the breakage but it's still going to touch a lot of places.

avi5079:
Hey @craigcook ,
is this issue still persist? if yes then i would like to work on it.

craigcook:
I think this is on hold while we work on migrating our docs to another platform. We're trying to avoid any major changes until we have our new doc site up and running (hopefully soon).

PR #1107, opened 2026-01-07
Refactor typography system to use CSS custom properties

stephaniehobson:
## Description

Refactor typography system to use CSS custom properties. This is the first PR in a refactor of how we use CSS variables (#982) and renaming `title` to `heading` (#668). Some kind of basic fallback support for older browsers will be added back in a latter PR once we can get a clear picture of how broken things are.

I know this looks big but for the most part it's just a straight find/replace on variable names or moving `@includes` to the end of declarations.

- Modernization:
  -  Default to CSS vars for font family, size, and line-height
    - Removed `@supports` declarations if they only had font declarations
    - `@include text-body-*` and `@include text-heading-*` mixins use CSS vars now 
  - Edited `@mixin font-size()` to stop outputting a pixel fallback
- Reorganization:
  - Renamed `-title-` to `-heading-` in mixins and CSS vars
    - and `mzp-u-heading-*` utility classes
    - component HTML/CSS classes will follow in a separate PR
  - Added font-family declaration to `@include text-body-*` mixins
  - Added `--theme-` prefix to variables expected to morph
  - Added `--token-` prefix to unchanging variables
    - Re-name font-size tokens to use a scale instead of tshirt sizes for names
  - Added `--theme-button-line-height` var
- Bug fixes:
  - Moved `@include text-*` to end of declarations

### Issue

#982 

### Testing

- See if it compiles
- Spot check for visual regressions
- Do not worry about older browser support fallbacks right now

---


claude.md
```diff
@@ -0,0 +1,45 @@
+# Protocol
```
stephaniehobson: Note to self: remove this.

PR #1120, opened 2026-01-27
Speed up CI and test in a matrix of Node 20–25

janbrasna:
## Description

The toolchain is already preinstalled on the runner, Jasmine is able to pick up all the browsers & drivers it's configured for. Removing the 1–2min of installation time, this enables expanding the runs to a matrix of Node versions (without wasting excessive resources), beyond what's used for publishing, to have some visibility and hints for consumers using different Node versions.

(With added caching the runs are changed to read-only.)

Also updates some prod docs deployment instructions.

### Issue

https://github.com/mozilla/bedrock/issues/16872

### Testing

https://github.com/janbrasna/protocol/actions/runs/21095848885 ✅


.github/workflows/node.js.yml
janbrasna, line 1: Hmmh git eventually decided there's too many changes along with the rename to track it as deletion and new addition instead of diffing the file:/ (I've renamed it from its GHA boilerplate defaults to actually match other uses to describe its content)… I see how this is not ideal, but I've made a diff on the side to hopefully make the changes more reviewable:

diff

janbrasna: FYI this also upgrades actions version to Node 24 compatible ones to avoid run annotation warnings added to results. _(ref https://github.com/mozilla/bedrock/issues/17092)_

janbrasna: There's a few things folded into this single PR to hopefully make it easier to ship the few changes in one go since it's pretty simple and short, however if it seems separating out the different bits (bumps, removal of preinstalled tools, rename, matrix strategy…) into distinct PRs would be beneficial I'm happy to split this up.

(Or if you want to see fresh CI runs since the logs have already expired, just lmk…)

stephaniehobson (approved):

.github/workflows/node.js.yml
janbrasna: For good measure CI run on the future runner: https://github.com/janbrasna/protocol/actions/runs/35923861825 ✅ 

_(Triggered this recently to check it still applies well to current main; and also, this is the diff 15daf44 I used for the above CI run, that makes this PR changeset a bit more readable if you actually wanna see the ± changes.)_

Issue #1121, opened 2026-01-29
Verify eslint-env inline configuration comments/globals are still necessary


janbrasna:
Before upgrading to ESLint v10 there will be a few warnings to look into. Apparently eslint-env comments in the webpack configs are still considered part of the eslintrc (non-flat) era:

> ESLintEnvWarning: `/* eslint-env */` comments are no longer recognized when linting with flat config and will be reported as errors as of v10.0.0. Replace them with `/* global */` comments or define globals in your config file. See https://eslint.org/docs/latest/use/configure/migration-guide#eslint-env-configuration-comments for details.

Quickly checking I _think_ these are already correctly configured in the right places from when Alex migrated to the new config style, so it may just be as simple as removing these remnants — but I'll cross-reference the values and environments first, to check it's actually 1:1 mapped already, or there are deviations worth mentioning.

PR #1129, opened 2026-02-28
Remove inline eslint globals in configs

janbrasna:
## Description

These are defined in the new config since #935 and should no longer use the old format.

### Issue

#1121

### Testing

CI pass on v9 should be enough to verify v10 compatibility of these removals.

### DIFF
--- a/webpack.package.static.config.js
+++ b/webpack.package.static.config.js
@@ -4,8 +4,6 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-/* eslint-env node, commonjs, es2017 */
-
 const CopyPlugin = require('copy-webpack-plugin');
 const path = require('path');
 const { CleanWebpackPlugin } = require('clean-webpack-plugin');
--- a/webpack.docs.build.config.js
+++ b/webpack.docs.build.config.js
@@ -4,8 +4,6 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-/* eslint-env node, commonjs, es2017 */
-
 const CssMinimizerPlugin = require('css-minimizer-webpack-plugin');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const RemoveEmptyScriptsPlugin = require('webpack-remove-empty-scripts');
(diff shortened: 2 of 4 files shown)


webpack.package.static.config.js
```diff
@@ -4,8 +4,6 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-/* eslint-env node, commonjs, es2017 */
-
 const CopyPlugin = require('copy-webpack-plugin');
 const path = require('path');
```
janbrasna, line 8: Worth noting the new globals are defined slightly differently:

https://github.com/mozilla/protocol/blob/d20837b673633ea2587e0290ce2e3354d32e1f79/eslint.config.js#L124-L132

janbrasna (commented):
The CI runs that twice:

1. [/actions/runs/22524118040/job/65253378571#step:5:33](https://github.com/mozilla/protocol/actions/runs/22524118040/job/65253378571#step:5:33)
2. [/actions/runs/22524118040/job/65253378571#step:6:10](https://github.com/mozilla/protocol/actions/runs/22524118040/job/65253378571#step:6:10)

and is apparently happy.

webpack.package.static.config.js
janbrasna: (But then again, whatever format the webpack configs become over time, the change from es-2017 to es-latest should have no meaningful impact.) — Expecting the newer flat config values having taken precedence, they've been used like this over a year anyways FWIW.

webpack.package.static.config.js
```diff
@@ -4,8 +4,6 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-/* eslint-env node, commonjs, es2017 */
-
```
janbrasna: Yeah not sure why GH doesn't see that as merge conflict by now, potentially by being the exact same change it's still happy to merge w/o conflict just resulting in an empty commit by now? Dunno. Removed in 9e37287 in a bigger update at this point. 🚀

janbrasna (commented):
Removed in #1133 already; closing.

PR #1135, opened 2026-09-09
Update browser support to Firefox ESR, Safari 15.6+, evergreen (Fix #1145)

stephaniehobson:
## Description

Update browser support matrix to Firefox ESR, Safari 15.6+, evergreen.

- Updates docs to reflect new support rules
- Update supported browser lists used by tools
- [x] I have documented this change in the design system.
- [x] I have recorded this change in `CHANGELOG.md`.

### Issue

Fix #1145

### Testing

Part of the v23 browser-support workstream (#1084). CSS/JS dead-code removal for the new matrix follows in separate stacked PRs.

PR #1137, opened 2026-09-09
Remove legacy browser-support CSS

stephaniehobson:
## Description

Removes some vendor prefixes. Unwraps some @supports declarations. Updates comments that are IE specific to explain they're legacy parts of the reset.

- [ ] ~I have documented this change in the design system.~
- [ ] ~I have recorded this change in `CHANGELOG.md`.~ (covered by existing messages)

### Issue

#1135


knowler (commented):
There are some details polyfill related styles that are left in https://github.com/mozilla/protocol/blob/2976625806a7ea026ff49a61d0b257dafba25d03/assets/sass/protocol/base/elements/_details.scss

stephaniehobson:
@knowler We should leave those. It's possible to invoke MzpDetails without using the details/summary elements and we want the styles between the two to match.

[There's a demo](https://protocol.mozilla.org/components/detail/details-component--custom) but it seems to be broken 🤦‍♀️. I made https://github.com/mozilla/protocol/issues/1158

knowler (approved):

PR #1143, opened 2026-09-10
Convert includes/ and base/ to CSS logical properties

stephaniehobson:
## Description

First pass of the logical-properties migration (goal 3/4 of #1084) — establishes the conversion pattern the rest of the workstream follows, and removes 23 of the 127 remaining `@include bidi()` calls.

Converts `includes/mixins/_utils.scss`, `includes/mixins/_details.scss`, `includes/forms/index.scss`, and `base/elements/_forms.scss`, `_lists.scss`, `_links.scss`, `_tables.scss`, `_quotes.scss`, `_details.scss`, and `base/utilities/_rich-text.scss`.

Every conversion verified against a before/after `sass` compile diff — confirmed each logical property produces byte-identical LTR and RTL output to the physical-plus-`[dir=rtl]`-override pair it replaced.

A couple of judgment calls worth a look:
- `select`'s `background-position` stayed physical with an explicit `[dir='rtl']` override — there's no standard logical keyword syntax for `background-position` safe across this matrix. Its accompanying padding tuple did convert cleanly to `padding-block`/`padding-inline`.
- Caught a real near-bug: moving `@include forms.form-input` to the end of the `select` rule would have silenced its last 2 `mixed-decls` warnings, but `form-input()` also sets a plain `padding` shorthand that would then win the cascade over my new logical padding, silently breaking the caret icon's clearance space. Reverted — those 2 warnings stay deferred, same as the mixed-decls PR concluded.

### Issue

Part of #1084

### Testing

`npm run lint`, `npm test` (47 specs, Firefox + Chrome), `npm run build-docs`, `npm run build-package`, and a before/after compile diff all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,11 @@ Introducing theme variables! CSS variables beginning with `--theme-` will adjust
   * Added status color variables (`--theme-color-success-*`, `--theme-color-error-*`, `--theme-color-warning-*`, `--theme-color-info-*`)
 * Removed Sass color variables from `_themes-sass.scss` (use CSS variables instead)
 
+### CSS Logical Properties
+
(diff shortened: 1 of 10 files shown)


stephaniehobson:
🙅‍♀️ Not ready for review - I found some RTL regressions.

PR #1159, opened 2026-09-21
Add viewport meta tag to fractal preview pages (Fix #1158)

### LINKED ISSUE #1158: Add viewport meta tag to fractal full page demos.
The [custom details component dem ](https://protocol.mozilla.org/components/preview/details-component--custom) requires a small viewport media query to trigger to work properly and right now it doesn't because there's no viewport declaration in the HTML.

stephaniehobson:
## Description

Add viewport meta tag to fractal preview pages

### Issue

Fix #1158

### Testing

[This page depends on media queries to trigger for the demo to work.](http://localhost:3000/components/preview/details-component--custom) At smaller sizes the text under the headings should collapse.


components/_preview-card-dark.html
```diff
@@ -2,6 +2,7 @@
 <html dir="ltr" class="no-js">
 <head>
   <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
```
knowler: We should be able to drop the `initial-scale=1` bit in these, but it doesn’t hurt to keep it (especially if it’s consistent elsewhere). It was necessary for older versions of iOS and Android, but has since been fixed. I only became aware of this within the last month or so.

```suggestion
  <meta name="viewport" content="width=device-width">
```

- https://www.matuzo.at/blog/2026/html-boilerplate#viewport-essential
- https://vale.rocks/micros/20260902-1350
  - In the thread, Declan did some testing and it was like iOS 8 and prior that needed it.

knowler (approved):

components/_preview-card-dark.html
stephaniehobson: Sweet! End of an era.

